PR #36, opened 2019-12-12
Deleting "Contributing" File.

deepak-vij:
As it is no longer needed. Wiki has the latest content for this topic. Otherwise, it would lead to duplication.


zasherif:
Not all developers accustomed to wikis, and subconsciously search for this file. So I'd rather just keep it.

deepak-vij:
In that case, we need to be aware of the duplicate content in two places. Going forward, let Wiki "Contributing" section be the master content for this. Anytime we make changes to it, we need to manually ensure that "Contribution/md" gets updated as well. Unless we figure out how to make the relative link work from wiki page to a file in repository.

At this time, I guess, we don't have any choice but to live with this duplicate content in two places.

deepak-vij:
I will go ahead and close this PR then.

PR #76, opened 2020-03-21
upload neutron plugin design

Eric-Yuan:


Documentation/design/neutron_plugin_for_mizar.md
```diff
@@ -0,0 +1,27 @@
+# Neutron plugin for mizar
+ 
+Openstack is one of the most popular opensource cloud platform, of which the network function subsystem is neutron. This Document is about to explan how to design a neutron plugin for mizar, in order that compute resources provided by openstack could use the mizar network pipeline. 
+
+neutron plugin might include two major functions: resource model save and transformation and network config distribution.
```
zasherif: I don't understand this statement. Is this what you want to say?

```
Mizar's neutron plugin includes two major functions:
1. Resource model translation from Neutron's network model to Mizar's network model
2. RPC Client calls to transit daemon to program Mizar's data-plane.
```

Documentation/design/neutron_plugin_for_mizar.md
```diff
@@ -0,0 +1,27 @@
...
+
+neutron plugin might include two major functions: resource model save and transformation and network config distribution.
+
+## Resource model save and transformation
+
+The plugin should provide all operation APIs for mizar resources(vpc, network, endpoint, etc.), as the resource models of mizar are not same as neutron(router, network, subnet, port, etc.). Besides, neutron is one of the core subsystems of openstack, other core subsystems like nova may couple with neutron models. In order to make other subsystems of openstack like nova could properly interaction with neutron, when to create a mizar resource we also need to create one or more neutron resources corrspondingly. Those co-created neutron resources should only be stored, as the original neutron agents may not be needed any more and the mizar agents(transit agent, etc.) might only concern about the mizar resources.
+
+![neutron plugin resource model map](./figures/neutron_plugin_model_map.png)
```
zasherif: The mapping should be:

Mizar -> Neuton
VPC -> Network
Network -> Subnet
Simple Endpoint -> Port

Documentation/design/neutron_plugin_for_mizar.md
```diff
@@ -0,0 +1,27 @@
...
+ 
+Openstack is one of the most popular opensource cloud platform, of which the network function subsystem is neutron. This Document is about to explan how to design a neutron plugin for mizar, in order that compute resources provided by openstack could use the mizar network pipeline. 
+
+neutron plugin might include two major functions: resource model save and transformation and network config distribution.
+
+## Resource model save and transformation
+
+The plugin should provide all operation APIs for mizar resources(vpc, network, endpoint, etc.), as the resource models of mizar are not same as neutron(router, network, subnet, port, etc.). Besides, neutron is one of the core subsystems of openstack, other core subsystems like nova may couple with neutron models. In order to make other subsystems of openstack like nova could properly interaction with neutron, when to create a mizar resource we also need to create one or more neutron resources corrspondingly. Those co-created neutron resources should only be stored, as the original neutron agents may not be needed any more and the mizar agents(transit agent, etc.) might only concern about the mizar resources.
```
zasherif: I'm not sure I fully understand the text. It will be helpful to provide examples and detailed workflows. This is how we design workflows in the new Mizar management plane: https://github.com/phudtran/mizar/blob/dev-k8s/Documentation/design/management_plane.md

Is this what you want to say?

```
1. The plugin shall transform all neutron's resource model's to Mizar resources
2. The Mizar resources shall be added to existing neutron network resources through API extension
3. The plugin shall program Mizar's transit daemon through RPC calls
4. ....
```

Documentation/design/neutron_plugin_for_mizar.md
```diff
@@ -0,0 +1,27 @@
...
+
+## Resource model save and transformation
+
+The plugin should provide all operation APIs for mizar resources(vpc, network, endpoint, etc.), as the resource models of mizar are not same as neutron(router, network, subnet, port, etc.). Besides, neutron is one of the core subsystems of openstack, other core subsystems like nova may couple with neutron models. In order to make other subsystems of openstack like nova could properly interaction with neutron, when to create a mizar resource we also need to create one or more neutron resources corrspondingly. Those co-created neutron resources should only be stored, as the original neutron agents may not be needed any more and the mizar agents(transit agent, etc.) might only concern about the mizar resources.
+
+![neutron plugin resource model map](./figures/neutron_plugin_model_map.png)
+
+## Network config distribution
+
+Because alcor could take the role of the controller of mizar, there might be two choices: joint with alcor or not.
```
zasherif: We don't need Alcor at all here, please remove the "joint with alcor" option

Documentation/design/neutron_plugin_for_mizar.md
```diff
@@ -0,0 +1,27 @@
...
+
+## Network config distribution
+
+Because alcor could take the role of the controller of mizar, there might be two choices: joint with alcor or not. 
+
+### Joint with alcor
+
+There would be a very thin driver for the plugin that it only need to send the resource operation to the alcor and receive the process result.
+
+![neutron plugin joint with alcor](./figures/neutron_plugin_joint_alcor.png)
```
zasherif: There is no value of doing this except adding extra overhead..

Documentation/design/neutron_plugin_for_mizar.md
```diff
@@ -0,0 +1,27 @@
...
+
+### Not joint with alcor
+
+The driver of the plugin should independently process the mizar resource operation, and send the network config directly to the transit agent/switch/router. Those process would be very similar with that in alcor or transit controller(mizar test codes).
```
zasherif: You send the configuration to the transit daemon and Not to the agent, switch, or router. These (switch/agent/router) are only resource extension to the neutron network model.

zasherif: We also need a detailed end-to-end workflow from neutron API server to transitd

Documentation/design/neutron_plugin_for_mizar.md
```diff
@@ -0,0 +1,27 @@
...
+
+There would be a very thin driver for the plugin that it only need to send the resource operation to the alcor and receive the process result.
+
+![neutron plugin joint with alcor](./figures/neutron_plugin_joint_alcor.png)
+
+### Not joint with alcor
+
+The driver of the plugin should independently process the mizar resource operation, and send the network config directly to the transit agent/switch/router. Those process would be very similar with that in alcor or transit controller(mizar test codes). 
+
+![neutron plugin not joint with alcor](./figures/neutron_plugin_not_joint_alcor.png)
```
phudtran: Moving forward, we have changed the names for the following resources.
Transit Switch -> Bouncer
Transit Router -> Divider

Documentation/design/neutron_plugin_for_mizar.md
Eric-Yuan: I get a little bit confused here. If a VPC mapping to a network means that all networks (Mizar) inside a VPC shared one boradcast zone? Only those crossing traffics between VPC need be sent to a  transit router (divider)?

Documentation/design/neutron_plugin_for_mizar.md
Eric-Yuan: OK

Documentation/design/neutron_plugin_for_mizar.md
Eric-Yuan: OK

Documentation/design/neutron_plugin_for_mizar.md
zasherif: Good question. Best explained by example.

Let's say we have a VPC with CIDR block 10.0.0.0/12 (This is a network in neutron)
We can have two networks below the VPCs net1: 10.1.0.0/24 and net2: 10.20.0.0/24 (These are subnets in neutron)

1. Traffic from net1 to net2 has to go through the divider
1. Traffic within net1 or within net2 does not go through the divider (only through the bouncer once and fast-path is then direct between endpoints)
1. If we have another VPC, then traffic between the two VPCs will go through the divider

As of broadcast/multicast traffic: we do not support them right now. We can discuss the real use cases in our next sprint planning. I would say broadcast/multicast may be supported within a network boundary (bouncer).

Note: Mizar data-plane does not work as traditional networks, for example, if we have an ARP request the bouncer does not broadcast it to all endpoints, because it knows where the target endpoint is. With this, a network is not necessarily a subnet within the VPC. It can be any arbitrary group of endpoints. 

Does that make sense? Again, we would love to discuss use cases of broadcast/multicast in the upcoming sprint planning.

zasherif:
@Eric-Yuan Quick follow-up here, are you planning to submit a revision addressing these comments soon?

Eric-Yuan:
> @Eric-Yuan Quick follow-up here, are you planning to submit a revision addressing these comments soon?

Yes, I will do it, sorry for the delay.

vinaykul:
This has become stale. We don't have neutron plugin on the roadmap. Closing this PR.

PR #138, opened 2020-04-09
Scaled Endpoint Delete

phudtran:


mgmt/common/workflow.py
```diff
@@ -102,10 +118,18 @@ def BouncerCreate(self, param):
...
 	def k8sServiceCreate(self, param):
 		pass
 
 	@abstractmethod
 	def k8sEndpointsUpdate(self, param):
 		pass
+
+	@abstractmethod
+	def k8sEndpointsDelete(self, param):
+		pass
```
zasherif, line 135: Do we still need this method?

mgmt/dp/mizar/workflows/bouncers/provisioned.py
```diff
@@ -42,5 +42,6 @@ def run(self):
 		bouncer = bouncers_opr.get_bouncer_stored_obj(self.param.name, self.param.spec)
 		bouncer.set_vni(vpcs_opr.store.get_vpc(bouncer.vpc).vni)
 		bouncer.droplet_obj = droplets_opr.store.get_droplet(bouncer.droplet)
+		bouncer.load_transit_xdp_pipeline_stage()
```
zasherif, line 45: +1

PR #145, opened 2020-04-12
Basic end to end tests (simple and scaled ep)

zasherif:
There are intermittent failing due to two existing bugs, to be solved next week.


teste2e/common/k8sservice.py
```diff
@@ -0,0 +1,11 @@
...
+from time import sleep
+
+SCRIPTS_DIR = '/var/mizar/test/scripts'
+
+class k8sService:
+
+    def __init__(self, api, name, ip):
+        self.api = api
+        self.name = name
+        self.ip = ip
```
phudtran, line 11: new line

PR #158, opened 2020-04-22
Maglev Consistent Hashing and Tests

phudtran:
Run the tests with `make mgmt_tests`


mizar/common/maglev_table.py
```diff
@@ -0,0 +1,141 @@
+import logging
+from zlib import crc32, adler32
+
+logging.basicConfig()
+logger = logging.getLogger()
+logger.setLevel(logging.INFO)
+
```
zasherif, line 7: We need to add a generic class here MaglevNode that we can extend with a Chain object. to remain functional in your test, you can have a __str__ method

zasherif (approved):
lgtm, one minor comment to change the element to a generic class

PR #165, opened 2020-04-24
Ftn operator

clu2xlu:


mizar/dp/mizar/operators/droplets/droplets_operator.py
```diff
@@ -99,6 +99,13 @@ def assign_divider_droplet(self, divider):
         d = random.sample(droplets, 1)[0]
         divider.set_droplet(d)
 
+    def assign_ftn_droplet(self, ftn):
+        droplets = set(self.store.get_all_droplets())
+        for d in droplets:
+            if not d.known_ftn:
+                ftn.set_droplet(d)
+                break
```
zasherif: How about moving this search loop to the store? Basically introduce a function get_non_ftn_droplets from the store to get a set of droplets that are not ftns. The store droplet apis need to maintain this list and return them in __constant time__. Then the assignment algorithm here selects one of them (random for now).

mizar/dp/mizar/operators/bouncers/bouncers_operator.py
```diff
@@ -56,7 +56,7 @@ def list_bouncers_obj_fn(name, spec, plurals):
             b = Bouncer(name, self.obj_api, self.store, spec)
             self.store_update(b)
 
-        kube_list_obj(self.obj_api, RESOURCES.droplets, list_bouncers_obj_fn)
+        kube_list_obj(self.obj_api, RESOURCES.bouncers, list_bouncers_obj_fn)
```
zasherif, line 59: Thanks

mizar/dp/mizar/operators/dividers/dividers_operator.py
```diff
@@ -56,7 +56,7 @@ def list_dividers_obj_fn(name, spec, plurals):
             if d.status == OBJ_STATUS.divider_status_provisioned:
                 self.store_update(d)
 
-        kube_list_obj(self.obj_api, RESOURCES.droplets, list_dividers_obj_fn)
+        kube_list_obj(self.obj_api, RESOURCES.dividers, list_dividers_obj_fn)
```
zasherif, line 59: Thanks

mizar/obj/droplet.py
```diff
@@ -43,6 +43,7 @@ def __init__(self, name, obj_api, opr_store, spec=None):
         self.known_bouncers = {}
         self.known_nets = {}
         self.known_eps = {}
+        self.known_ftn = {}
```
zasherif: See my comment on using the store object instead. All what you need then is to have one field 
```
self.is_ftn
```

mizar/obj/droplet.py
```diff
@@ -150,3 +151,14 @@ def delete_ep(self, name, ep):
             self.known_eps.pop(name)
             if ep.ip not in self.known_eps.values():
                 self.rpc.delete_ep(ep)
+
+    def update_dft(self, ftn):
+        if not self.known_ftn:
+            self.known_ftn[ftn.name] = ftn.dft
+        #self.rpc.update_dft(ftn)
```
zasherif: ditto

zasherif (changes requested):

PR #169, opened 2020-05-19
CNI version fix

phudtran:


mizar/daemon/app.py
```diff
@@ -37,7 +37,7 @@
 
 # Setup the droplet's host
 script = (f''' bash -c '\
-nsenter -t 1 -m -u -n -i rm /etc/cni/net.d/10-kindnet.conflist &&\
+nsenter -t 1 -m -u -n -i rm -rf /etc/cni/net.d/10-kindnet.conflist && \
```
zasherif: This is kind specific. Can we do a similar thing to remove existing conflist except mizar conf

zasherif (approved):
A minor comment, ow lgtm

mizar/daemon/app.py
phudtran: 👍  Fixed

zasherif (approved):

PR #171, opened 2020-05-23
Add a file about gateway node

HuaqingTu:
This document introduces the model that flow rules are distributed on demand based on gateway nodes.


zasherif (changes requested):
What is this?! Would you please give more context and follow the contributing guideline?
Please note that none of the core engineers here understands Chinese, so please make all your design proposals in __English__. Otherwise, we are not able to review it, change it, or give you any feedback. Also, all the design proposals in Mizar are written in __Markdown__, so we can collaboratively provide comments and introduce changes later in git. Would you please write your proposals in markdown, and in English?

PR #173, opened 2020-06-01
Design doc for arktos integration

clu2xlu:


docs/design/mp_arktos.md
```diff
@@ -0,0 +1,145 @@
...
+###	Pod creation
+* Customer creates a new pod using command: 
+```
+kubectl run pod1 --image=localhost:5000/testpod
+```
+* Then this newly created pod is placed on one of the host/droplet. 
+* Customer has the choice of associating this pod with a service using label flag:
+```
+kubectl label pods pod1 run=example3 –overwrite
+```
```
zasherif: How does arktos pass the VPC/Network information in Pod creation?

docs/design/mp_arktos.md
```diff
@@ -0,0 +1,145 @@
...
+        service.beta.kubernetes.io/mizar-scaled-endpoint-type: “scaled-endpoint”
+  lables: 
+    run: test-service-3
+spec:
+  ports: 
+-	port: 80
+     protocol: TCP
+  selector: 
+     run: example3
+```
```
zasherif: How does arktos pass vpc/net info here?

docs/design/mp_arktos.md
```diff
@@ -0,0 +1,145 @@
...
+
+## Mizar VPC Object Overview: 
+In Mizar, VPC object has two required fields: ip and prefix, where in combination, they compose the CIDR of the VPC. For example, for CIDR 10.0.0.0/16, the ip = 10.0.0.0 and prefix = 16. 
+```yaml
+spec:
+  ip: 10.0.0.0
+  prefix: 16
+```
+## Proposal 
+Given the fact that Arktos network object is equivalent to vpc object in Mizar, I would like to propose two options:
```
zasherif, line 48: In both option1 and option2, can you give more details how are we going to create a mizar network within the VPC? I remember you mentioned passing multiple CIDR blocks.

docs/design/mp_arktos.md
```diff
@@ -0,0 +1,145 @@
+<!--
+SPDX-License-Identifier: MIT
+Copyright (c) 2020 The Authors.
+
+Authors: Sherif Abdelwahab <@zasherif>
+         Phu Tran          <@phudtran>
+         Catherine Lu      <@clu2>
```
zasherif: Please keep only your name here.

zasherif (changes requested):

docs/design/mp_arktos.md
clu2xlu: That's when we ask Arktos team to provide VPC CIDR, and they said they can provide subnet CIDR in the annotation section like this: 

type: Network
metadata:
  name: network-foo
  annotations:
     subnets: ["192.168.0.0/24", "192.168.1.0/24"]
spec:
  type: mizar  // vpc is not a key; put mizar or the value that network provider recognizes
  vpcID: mizar-12345

But above format will not work since we still need vpc CIDR range. Two subnet CIDR's do not tell us the CIDR of the VPC.

docs/design/mp_arktos.md
```diff
@@ -0,0 +1,172 @@
...
+metadata:
+  name: nginx
+spec:
+  containers:
+  - name: nginx
+    image: nginx
+    ports:
+      - containerPort: 443
+  network: vpc-1
+  nics:
```
XiaoningDing, line 162: If the IP is or subnet is specified by user, which component will assign them?

clu2xlu: First, we should check if the IP address is within subnet's CIDR range. If true, then the ip will be the pod's ip address, and subnet name listed will be the subnet of the pod.

docs/design/mp_arktos.md
```diff
@@ -0,0 +1,172 @@
...
+substantial portions of the Software.THE SOFTWARE IS PROVIDED "AS IS",
+WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED
+TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE
+FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR
+THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+-->
+
+# Arktos Network and Mizar Integration
```
XiaoningDing, line 23: We also need to address the workflow of network object deletion. For example, do we always delete the VPC is a network object is deleted (even if the network objects references an existing VPC ID?)  Do we need to add a finalizer on the network object? etc.  Also I guess computing side probably should guarantee a network object can only be deleted when all referencing entities(pods, services,...) are deleted.

docs/design/mp_arktos.md
XiaoningDing: I'm sorry - I missed a keyword.  What I wanted to say is, "if the IP or subnet is **not** specified by user, which component will assign them?"

docs/design/mp_arktos.md
XiaoningDing: And do we support cases of updating network object?

clu2xlu: Yes, it will be the same workflow as create. i.e. create workflow supports both create and update

docs/design/mp_arktos.md
clu2xlu: If neither IP or subnet is specified, Mizar will pick the first subnet within ```vpc-1``` and assign pod to that subnet.

docs/design/mp_arktos.md
```diff
@@ -0,0 +1,172 @@
...
+	```self.store.vpcs_arktosnet_store[arktosnet.name] = {}```
+    ```self.store.vpcs_arktosnet_store[arktosnet.name][vpc.name] = vpc_object```
+
+The Arktos network object is associated with services in similar ways: 
+    ```self.store.services_arktosnet_store[arktosnet.name] = {}```
+    ```self.store.services_arktosnet_store[arktosnet.name][service.name] = service_object```
+
+The subnet object is associated with pods as: 
+    ```self.store.pods_net_store[net.name] = {}```
+    ```self.store.pods_net_store[net.name][pod.name] = pod_object```
```
phudtran: Do we need a new "pod" object here?
Can't we just use the preexisting endpoint object?

docs/design/mp_arktos.md
```diff
@@ -0,0 +1,172 @@
...
+This document is designated for Arktos service using Mizar as its underline network system. In this document, you will find the design detail on how to integration Arktos multi-tenancy network object with Mizar as its network platform, so that Arktos can fully utilize network objects offered in Mizar, including vpcs, nets, services, pods, bouncers and dividers. 
+
+## Arktos Network Object Overview: 
+The Arktos network object contains its type and type-specific configurations, which translates into spec fields type and vpcID:
+```yaml
+spec:
+  type: mizar
+  vpcID: vpc-1a2b3c4d
+```
+The type field is the only required field, where it identifies the underline network service (which in our case is mizar). VpcID field is optional, and it lists the vpc id that is associated with this Arktos network object. If vpcID is not specified, the default VPC will be used.
```
phudtran, line 36: What if multiple Arktos Network objects are created with no VpcID field?

phudtran (changes requested):

docs/design/mp_arktos.md
clu2xlu: Correct, pod = ep. Not need to have pods_net_store. It should be our existing eps_net_store

docs/design/mp_arktos.md
h-w-chen: in short, CNI_ARGS as in cni spec.
please see [annotation as cni-ags hints](https://github.com/futurewei-cloud/arktos/pull/341) for more description.

docs/design/mp_arktos.md
h-w-chen: networks inside mizar has separate life cycle than network objects in arktos. For now, arktos puts network object in Terminating phase if it is requested to delete, and finalizers prevents its deletion until they are coped with - ideally by appropriate garbage collection controllers. What are the cleanups has not been decided yet.

h-w-chen (commented):
please look at [arktos pod-from-creation-to-running](https://github.com/futurewei-cloud/arktos/pull/341) to see whether it has addressed the concerns in this PR.

docs/design/mp_arktos.md
h-w-chen: on pod creation, network controller is supposed to extract vpc etc info from pod label or annotations, and do proper network resource preparation; 
on cni invocation, CNI_ARGS passed as in cni spec.
please see [annotation as cni-ags hints](https://github.com/futurewei-cloud/arktos/pull/341) for more description.

zasherif (changes requested):
Is this document up-to-date? We need to reflect the latest changes!

PR #175, opened 2020-06-08
Arktos and Mizar Integration

clu2xlu:


docs/design/mp_arktos.md
```diff
@@ -0,0 +1,229 @@
...
+    arktos.futurewei.com/network: vpc-1
+spec:
+  containers:
+  - name: nginx
+    image: nginx
+    ports:
+      - containerPort: 443
+```
+
+When a pod is attached to a certain network, then it needs to specify subnet or IP. It can be expressed in annotations:
```
phudtran: Do you mean "attached to a certain **subnet**" here?

docs/design/mp_arktos.md
```diff
@@ -0,0 +1,229 @@
...
+
+Annotation ```arktos.futurewei.com/nic``` is for user to provide optional information about pod nic. The recognized keys of element include:
+
+|key|exemplary value|
+|---:|---:|
+|name|eth0|
+|subnet|subnet-1, or 192.168.100.0/26|
+|ip|192.168.100.5|
+
+Please note that any of the three fields can be missing. For example, user might only specify  ```ip``` field; in this case, Mizar needs to figure out which subnet this ip falls into, and creates the pod into the correct subnet.
```
phudtran: Will a subnet be created if no pre-existing subnet contains the IP address in its range?

docs/design/mp_arktos.md
```diff
@@ -0,0 +1,229 @@
...
+  network: my-network
+  selector:
+    app: MyApp
+  ports:
+    - protocol: TCP
+      port: 80
+      targetPort: 9376
+```
+* Create ```my-service```
+* Update object store: map Arktos network with services: ```self.store.services_arktosnet_store[my-network][my-service] = service_object```
```
phudtran: Will we be provisioning a default number of backends for this service/scaled ep?
Will we allow the user to specify how many backends they desire for their service/scaled ep?

phudtran (changes requested):

docs/design/mp_arktos.md
clu2xlu: "network" here is arktos network object. Will make the change

docs/design/mp_arktos.md
clu2xlu: For version one, we do not support any new object creation on Mizar side. If no pre-existing subnet, mizar will throw an error

docs/design/mp_arktos.md
clu2xlu: This part I need more detail from arktos team on how services are managed.

docs/design/mp_arktos.md
```diff
@@ -0,0 +1,229 @@
...
+```yaml
+apiVersion: arktos.futurewei.com/v1
+kind: Network
+metadata:
+  name: vpc-1
+spec:
+  type: mizar
+  vpcID: vpc-1a2b3c4d
+```
+The type field is the only required field, where it identifies the underline network service (which in our case is mizar). VpcID field is optional, and it lists the vpc id that is associated with this Arktos network object. If vpcID is not specified, the default VPC will be used.
```
h-w-chen: in mizar case, vpcID is required, I think.

docs/design/mp_arktos.md
```diff
@@ -0,0 +1,229 @@
...
+kind: Network
+metadata:
+  name: vpc-1
+spec:
+  type: mizar
+  vpcID: vpc-1a2b3c4d
+```
+The type field is the only required field, where it identifies the underline network service (which in our case is mizar). VpcID field is optional, and it lists the vpc id that is associated with this Arktos network object. If vpcID is not specified, the default VPC will be used. 
+
+Below is the definition of a flat network, where the defaul network is used:
```
h-w-chen: flat network is not the concern of mizar. maybe you want remove flat network part.

docs/design/mp_arktos.md
h-w-chen: specifying subnet/ip is totally optional.

clu2xlu: yes, that's correct. I will change the wording so it's more clear

docs/design/mp_arktos.md
```diff
@@ -0,0 +1,229 @@
...
+    image: nginx
+    ports:
+      - containerPort: 443
+```
+Operator kicks off creation process:
+* List Kubernetes cluster object / Arktos newtork object
+* Retrieve ```labels``` and ```annotations``` information: 
+    * ```arktos.futurewei.com/network```: vpc-1
+    * ```arktos.futurewei.com/nic```: {"name": "eth0", "ip": "10.10.1.12", "subnet": "net1"}
+* Find Arktos network object that is listed in ```arktos.futurewei.com/network```
```
h-w-chen, line 190: suggest to use mizar.futurewei.com/network-user-input, as this annotation is mizar specific.

docs/design/mp_arktos.md
```diff
@@ -0,0 +1,229 @@
...
+* Find subnet and ip information from ```arktos.futurewei.com/nic```, and choose the correct subnet to place the new pod. 
+* If neither IP nor subnet is specified, Mizar will pick the first subnet within ```vpc-1``` and assign pod to that subnet.
+* Update object store: map subnet object (which is equivalent to net object in Mizar) with pod: ```self.store.eps_net_store[net1][nginx] = pod object```
+* After Pod creation is completed, update Pod object with new annotations: ```arktos.futurewei.com/network_user_input``` and ```arktos.futurewei.com/network-readiness```.
```
h-w-chen: suggest to use mizar.futurewei.com/network-user-input, as this annotation is mizar specific.

docs/design/mp_arktos.md
```diff
@@ -0,0 +1,229 @@
...
+* List Kubernetes cluster object / Arktos newtork object
+* In case of using services, vpcID should not listed, and default network is selected
+* Waiting on customer to create service using yaml file:  
+```yaml
+apiVersion: v1
+kind: Service
+metadata:
+  name: my-service
+spec:
+  network: my-network
```
h-w-chen: should use label instead. I knew this was copied from arktos doc, which has inconsistency problem.

h-w-chen (commented):
Can this doc elaborate a bit about:
1. is there any special requirements that cni plugin need to meet? 
2. how do pods access k8s api-servers by their node ip addresses?
3. how is service ip address assigned? is it assigned by mizar, or expecting arktos to assign it?

clu2xlu:
> Can this doc elaborate a bit about:
> 
> 1. is there any special requirements that cni plugin need to meet?
> 2. how do pods access k8s api-servers by their node ip addresses?
> 3. how is service ip address assigned? is it assigned by mizar, or expecting arktos to assign it?

addressed in latest commit

docs/design/mp_arktos.md
```diff
@@ -0,0 +1,241 @@
...
+                    "interface": 0
+                }
+            ]
+        }
+
+    def get_pod_obj(self, params):
+        logger.debug("Allocate endpoint name")
+        name = ""
+        if 'K8S_POD_NAME' in params.cni_args_dict:
+            name = params.cni_args_dict['K8S_POD_NAME']
```
h-w-chen, line 193: it seems mizar will use pod-name as the only key. This works fine most of the cases; however it might conflict with other namespace/tenant.

docs/design/mp_arktos.md
```diff
@@ -0,0 +1,241 @@
...
+  selector:
+    app: MyApp
+  ports:
+    - protocol: TCP
+      port: 80
+      targetPort: 9376
+```
+* Note that if ```arktos.futurewei.com/network``` not listed, and default network is selected
+* Build-in operator kicks off ```my-service``` creation
+* Use endpoint operator to create a scaled endpoint object within ```vpc-1```.
```
h-w-chen, line 238: it seems the service ip is decided by mizar; if so, please update spec.ClusterIP it in arktos service object.

clu2xlu: No problem, this is something we already have in our current design.

docs/design/mp_arktos.md
clu2xlu: I am thinking to use K8S_POD_NAMESPACE as a way to determine the tenant ID. However, since metadata.tenant and metadata.namespace is different in Arktos, what is the best way for Mizar to get tenant information? As I mentioned during our conversation off-line, the arktos network object definition and pod definition do not have tenant information.

clu2xlu: Is it possible for two different customers/tenants to create arktos network objects with the same name? If yes, we should re-evaluate object store association. 
Also, is it possible for two tenants have the same name?

h-w-chen: in theory it is possible.
tenants won't be the same; tenant name is unique.

docs/design/mp_arktos.md
```diff
@@ -155,22 +146,28 @@ spec:
...
 Operator kicks off creation process:
 * List Kubernetes cluster object / Arktos newtork object
-* Retrieve ```labels``` and ```annotations``` information: 
+* Retrieve metadata information: ```metadata.tenant```, ```metadata.namespace```, ```metadata.labels``` and ```metadata.annotations```: 
+    * ```tenant```: customerA
+    * ```namespace```: internal
     * ```arktos.futurewei.com/network```: vpc-1
     * ```arktos.futurewei.com/nic```: {"name": "eth0", "ip": "10.10.1.12", "subnet": "net1"}
+* Get container id of the pod, i.e. ```status.containerStatuses[0].containerID```
+* Set pod name as: ```{name}```-```{status.containerStatuses[0].containerID[-9:]}``` (use last 9 characters of container id to differentiate containers with the same ```name``` field)
```
h-w-chen: not sure there is container-id available at this moment.

h-w-chen (approved):
from the design view of interaction with Arktos, this PR looks Okay to me.
The critical information is key derived from {tenant + ns + pod name}. One solution is explicit passage by leveraging ani-args annotation; another possible way is implicit env var of **K8S_POD_TENANT** (seems already provided by Arktos) which will be set by arktos just like CNI_POD_NAME  .

I approve this design at aspect of arktos interaction.

clu2xlu:
Prefer env var of CNI_TENAME_NAME approach.

docs/design/mp_arktos.md
h-w-chen: Arktos provides K8S_POD_TENANT along with K8S_POD_NAMESPACE & K8S_POD_NAME.
Consider to leverage that if you would like to use tenant+ns+pod as key and seek for alternative that does not rely on explicit ani-args mechanism.

phudtran (approved):

PR #183, opened 2020-07-08
New CNI, interface and droplet services

zasherif:


mizar/daemon/app.py
```diff
@@ -14,13 +15,48 @@
 import sys
 sys.path.append("/home/sherif/cloudfabric/dev/mizar")
```
clu2xlu, line 16: Should be /home/ubuntu/mizar

zasherif: Forgotten from manual testing. my bad. We don't actually need this.

phudtran (approved):
LGTM since we went through it together.

PR #185, opened 2020-07-10
Add CNI/Interface delete action

clu2xlu:


mizar/cni/mizarcni.py
```diff
@@ -165,12 +168,31 @@ def activate_interface(self, interface):
...
+            iproute_ns.remove()
+        exit(0)
 
     def do_get(self):
         logger.info("CNI get is not implemented")
         print("")
+        val = "!!get service!!"
+        status = 1
+        logger.info("cni service get {}".format(params))
+        return val, status
```
zasherif: We don't need this code. Right?

mizar/daemon/interface_service.py
```diff
@@ -34,6 +34,7 @@ def __init__(self):
         self.iproute = IPRoute()
         self.interfaces = {}  # In-memory tracking for Pod interfaces
         self.queued_pods = set()  # A set of pods, with queued interfaces (to be consumed)
+        self.dequeued_pods = set()  # A set of pods, with queued interfaces (to be deleted)
```
zasherif: Do we actually need this? In the create we need a kind of synchronization between the operator and the cni, that's why we have the queue. But for delete, you can simply delete the interface in the CNI and delete the simple endpoint in the operator.

mizar/daemon/interface_service.py
```diff
@@ -107,6 +108,14 @@ def ProduceInterfaces(self, request, context):
...
         return interfaces
 
+    def RemoveInterfaces(self, request, context):
+        """
+        """
+        interfaces = request
+        for interface in interfaces.interfaces:
+            self._DequeueInterface(interface)
+        return interfaces
+
```
zasherif: I don't believe we need this RPC at all.

mizar/daemon/interface_service.py
```diff
@@ -165,6 +204,16 @@ def _ProvisionVethInterface(self, interface, cni_params):
...
+    def _DeleteVethInterface(self, interface, cni_params):
+        """
+        Delete a veth interface
+        """
+        veth_peer_index = get_iface_index(interface.veth.peer, self.iproute)
+        self.iproute.link('del', index=veth_peer_index)
+
+        # Update the Transit Agent
+        self._DeleteTransitAgent(interface)
+
```
zasherif: This is pretty much what you need!

mizar/cni/mizarcni.py
```diff
@@ -165,12 +168,31 @@ def activate_interface(self, interface):
...
+
+        for interface in interfaces:
+            veth_index = get_iface_index(interface.veth.name, self.iproute)
+            self.iproute.link('del', index=veth_index)
+            lo_idx = get_iface_index("lo", self.iproute)
+            self.iproute.link('del', index=lo_idx)
+            head, netns = os.path.split(self.netns)
+            iproute_ns = NetNS(netns)
+            iproute_ns.close()
+            iproute_ns.remove()
```
zasherif: Isn't this redundant to ``` def _DeleteVethInterface(self, interface, cni_params) ``` ?

mizar/daemon/interface_service.py
```diff
@@ -126,6 +135,21 @@ def _QueueInterface(self, interface):
...
         interface.status = InterfaceStatus.queued
 
+    def _DequeueInterface(self, interface):
+        pod_name = get_pod_name(interface.interface_id.pod_id)
+        logger.info("Removing interface {}".format(interface))
+        with self.interfaces_lock:
+            # Append the interface to the pod's interfaces (important in
+            # multi-interfaces case)
+            if pod_name in self.interfaces:
+                self.interfaces[pod_name].remove(interface)
```
zasherif: You can simply move this to ```DeleteInterface```, everything else seems like not needed.

mizar/daemon/interface_service.py
```diff
@@ -179,6 +228,16 @@ def _ConfigureTransitAgent(self, interface):
         self.rpc.update_agent_metadata(interface)
         self.rpc.update_ep(interface)
 
+    def _DeleteTransitAgent(self, interface):
+        self.rpc.load_transit_agent_xdp(interface)
```
zasherif: Why are you loading the transit agent after deleting the interface? This is guaranteed to fail. I think you meant an unload transit agent. Actually, once you delete the interface, unloading is probably not needed. Can you verify that the XDP program (transit Agent) will not remain after you delete the interface?

mizar/daemon/interface_service.py
```diff
@@ -179,6 +228,16 @@ def _ConfigureTransitAgent(self, interface):
...
+
+        for bouncer in interface.bouncers:
+            self.rpc.delete_agent_substrate_ep(
+                interface.veth.peer, bouncer.ip_address)
+
+        self.rpc.update_agent_metadata(interface)
+        self.rpc.update_ep(interface)
```
zasherif: Not needed at all in delete.

mizar/daemon/interface_service.py
```diff
@@ -179,6 +228,16 @@ def _ConfigureTransitAgent(self, interface):
         self.rpc.update_agent_metadata(interface)
         self.rpc.update_ep(interface)
 
+    def _DeleteTransitAgent(self, interface):
+        self.rpc.load_transit_agent_xdp(interface)
+
+        for bouncer in interface.bouncers:
+            self.rpc.delete_agent_substrate_ep(
+                interface.veth.peer, bouncer.ip_address)
```
zasherif: Well, this can only happen from the operator. Because we need to maintain a reference count of the usage of this substrate endpoint! You might still keep the ref count for the substrate in this interface service as well. I think that shall work.

mizar/proto/mizar/proto/interface.proto
```diff
@@ -36,6 +35,8 @@ enum InterfaceStatus {
   init = 0;
   queued = 1;
   consumed = 2;
+  dequeued = 3; 
+  deleted = 4;
```
zasherif: Since you clean-up, the objects from the interfaces table, Do we really need these status values?

zasherif (changes requested):

mizar/cni/mizarcni.py
```diff
@@ -141,14 +141,17 @@ def activate_interface(self, interface):
...
+        iproute_ns = NetNS(netns)
+        veth_index = get_iface_index(interface.veth.name, self.iproute)
 
-        # TODO (cathy): skip the interface activation if an interface with same
+        # skip the interface activation if an interface with same
         # interface.interface_id.interface already UP and configured in the
         # namespace
+        actived_idxs = iproute_ns.link_lookup(operstate="UP")
+        if (veth_index in actived_idxs)
+            exit(0)
```
zasherif: We cannot exit here, there might be other interfaces to activate. Right?

zasherif (changes requested):

mizar/daemon/interface_service.py
clu2xlu: For delete, I thought we still need queue all the interface, and then delete one by one, just like add.

zasherif: For ADD, there is the information passed by the operator (IP address, ...etc). For DEL, do we have such a case with different interface related configurations in CNI and in Operator? I don't think so. Primarily the DEL command is self-contained, and you shall be able to delete the interface entirely in the CNI without waiting for the operator. If there are ref-counting (like the substrate) and you need to maintain that ref-count in the operator, you can make an independent call to manage that ref-count, but I really don't see that even needed.

mizar/cni/mizarcni.py
```diff
@@ -141,32 +141,38 @@ def activate_interface(self, interface):
...
-        lo_idx = iproute_ns.link_lookup(ifname="lo")[0]
-        iproute_ns.link('set', index=lo_idx, state='up')
+        # skip the interface activation if an interface with same
+        # interface.interface_id.interface already UP and configured in the
+        # namespace
+        actived_idxs = iproute_ns.link_lookup(operstate="UP")
+        if (veth_index not in actived_idxs):
+            logger.error("Move interface {}/{} to netns {}".format(
+                interface.veth.name, veth_index, netns))
+            self.iproute.link('set', index=veth_index, net_ns_fd=netns)
```
zasherif: If the interface is already in the netns (as you got in actived_idxs), why moving the interface again to the netns?

The function needs to return here since this implies that the interface is already in the netns (and been configured before).

mizar/cni/mizarcni.py
```diff
@@ -141,32 +141,38 @@ def activate_interface(self, interface):
...
+
+            # configure and activate interfaces inside the netns
+            iproute_ns.link('set', index=veth_index, ifname=self.interface)
+            lo_idx = iproute_ns.link_lookup(ifname="lo")[0]
+            iproute_ns.link('set', index=lo_idx, state='up')
```
zasherif: Why we do need this here? If the function does not return, the configuration will be done again after the branching. Right?

mizar/daemon/interface_service.py
```diff
@@ -219,10 +227,23 @@ def ConsumeInterfaces(self, request, context):
...
+        """
+        Delete network interfaces for a pod
+        """
+        # input is pod_id 
 
+        cni_params = request
+        pod_name = get_pod_name(cni_params.pod_id)
+        interfaces = self.interfaces.get(pod_name, [])
+        logger.debug("Deleting interfaces for pod {}".format(pod_name))
+        print(interfaces)
```
zasherif: Please remove this print.

mizar/daemon/interface_service.py
```diff
@@ -219,10 +227,23 @@ def ConsumeInterfaces(self, request, context):
...
         return self._ConsumeInterfaces(requested_pod_name, cni_params)
 
-    def DeleteInterface(self, request, context):
-        # TODO (Cathy): implement the reverse logic
-        return empty_pb2.Empty()
+    def DeleteInterfaces(self, request, context):
+        """
+        Delete network interfaces for a pod
+        """
+        # input is pod_id
```
zasherif: Please remove this comment.

mizar/daemon/interface_service.py
```diff
@@ -144,6 +135,14 @@ def _ConsumeInterfaces(self, pod_name, cni_params):
         logger.info("Consumed {}".format(interfaces))
         return interfaces
 
+    def _DeleteInterfaces(self, pod_name, cni_params):
+        with self.interfaces_lock:
+            interfaces = self.interfaces.get(pod_name, [])
+            for interface in interfaces:
+                if interface.interface_type == InterfaceType.veth:
+                    self._DeleteVethInterface(interface, cni_params)
+                    logger.info("Removed {}".format(interface))
```
zasherif: I believe this is problematic. This loop will delete all the interfaces for the Pod. With one interface scenario, that's fine. With multiple interfaces, this function must only delete the interface requested by the CNI.

mizar/daemon/interface_service.py
```diff
@@ -219,10 +227,23 @@ def ConsumeInterfaces(self, request, context):
...
 
+        cni_params = request
+        pod_name = get_pod_name(cni_params.pod_id)
+        interfaces = self.interfaces.get(pod_name, [])
+        logger.debug("Deleting interfaces for pod {}".format(pod_name))
+        print(interfaces)
+        for interface in interfaces:
+            if pod_name in self.interfaces:
+                self.interfaces[pod_name].remove(interface)
+            self._DeleteInterfaces(pod_name, interface.interface_id)
```
zasherif: The function signatutre takes: pod_name, cni_paramter. Here it passed interface_id as the second parameter. Then the function itself does not make any use of the second parameter. I believe the _DeleteInterfaces function must delete only the interface that the CNI requests. Would you please fix this?

zasherif (changes requested):
This still needs a lot of work. Two things stand out for me:
- The built-in POD operator changes needs to go with this PR to see the end-to-end flow
- There must be a ref-count for the substrate endpoint and delete it when not used by the any interface on the host (either in Interface Service or in the Pod operator). Please check this part with @phudtran.

mizar/cni/mizarcni.py
clu2xlu: here veth_index is NOT in actived_idxs. So it will still need follow the activation steps

zasherif: Oh! I did not notice that. This branch shall check if the interface is activated and do nothing. What is the point of checking it is not activated, if the function will activate it anyway after branching?

phudtran (commented):
I've implemented the reference counting for delete previously, but since the rpc's are now made by the CNI, the calls are not being tracked.
See [droplet.py](https://github.com/futurewei-cloud/mizar/blob/dev-next/mizar/obj/droplet.py#L147)

mizar/cni/mizarcni.py
```diff
@@ -142,13 +142,14 @@ def activate_interface(self, interface):
...
-        # interface.interface_id.interface already UP and configured in the
-        # namespace
-
         head, netns = os.path.split(self.netns)
         iproute_ns = NetNS(netns)
         veth_index = get_iface_index(interface.veth.name, self.iproute)
+
+        actived_idxs = iproute_ns.link_lookup(operstate="UP")
+        if (veth_index in actived_idxs):
+            return
```
zasherif, line 151: Thanks :)

mizar/cni/mizarcni_operator_test.py
```diff
@@ -0,0 +1,82 @@
...
+
+interfaces = InterfacesList(interfaces=[inputInterface])
+
+interface = InterfaceServiceClient("localhost").ProduceInterfaces(interfaces)
+
+print("ProduceInterfaces Input: {}".format(interface))
+
+print("DeleteInterfaces input: {}".format(interfaces))
+
+InterfaceServiceClient("localhost").DeleteInterfaces(interfaces.interfaces[0].interface_id)
```
zasherif, line 82: You can delete this file. I already removed it from the merged code in dev-next

mizar/common/workflow.py
```diff
@@ -143,3 +143,7 @@ def k8sDropletCreate(self, param):
     @abstractmethod
     def k8sPodCreate(self, param):
         pass
+    
+    @abstractmethod
+    def k8sPodDelete(self, param):
+        pass
```
zasherif, line 149: new line

mizar/daemon/interface_service.py
```diff
@@ -219,8 +221,51 @@ def ConsumeInterfaces(self, request, context):
...
+        if obj_name in self.known_substrates.keys():
+            self.known_substrates.pop(obj_name)
+        self.rpc.update_substrate_ep(droplet_ip, droplet_mac)
+
+    def _DeleteVethInterface(self, interface):         
+        """
+        Delete a veth interface
+        """
+        veth_peer_index = get_iface_index(interface.veth.peer, self.iproute)
+
```
zasherif, line 248: extra line

mizar/daemon/interface_service.py
```diff
@@ -219,8 +221,51 @@ def ConsumeInterfaces(self, request, context):
...
+        self.rpc.update_substrate_ep(droplet_ip, droplet_mac)
+
+    def _DeleteVethInterface(self, interface):         
+        """
+        Delete a veth interface
+        """
+        veth_peer_index = get_iface_index(interface.veth.peer, self.iproute)
+        
+        self.iproute.link('del', index=veth_peer_index)
+
```
zasherif, line 250: ditto

mizar/dp/mizar/workflows/builtins/pods/delete.py
```diff
@@ -24,3 +24,35 @@
...
+    def run(self):
+        logger.info("Run {task}".format(task=self.__class__.__name__))
+        eps = endpoints_opr.store.get_eps_in_pod(self.param.name)
+        logger.info("Deleting Pod {}".format(self.param.name))
+
+        for ep in eps:
+            ep_obj = endpoints_opr.store.get_ep(ep)
+            endpoints_opr.delete_simple_endpoint(ep_obj)
+
+        self.finalize()
```
zasherif, line 58: new line

mizar/store/operator_store.py
```diff
@@ -128,6 +133,14 @@ def delete_ep(self, name):
...
         if l == 0:
             del self.eps_net_store[ep.net]
+        
+        if ep.k8s_pod_name in self.eps_pod_store.keys():
+            if name not in self.eps_pod_store[ep.k8s_pod_name]:
+                return ep
+            self.eps_pod_store[ep.k8s_pod_name].pop(name)
+            l = len(self.eps_pod_store[ep.k8s_pod_name])
+            if l == 0:
+                del self.eps_pod_store[ep.k8s_pod_name]
```
zasherif, line 143: What is the use of that block? Why do we even need an eps_pod_store?

mizar/obj/tests/test_arktos_pod.yaml
```diff
@@ -0,0 +1,14 @@
...
+  labels:
+    arktos.futurewei.com/network: vpc0
+  annotations:
+    arktos.futurewei.com/nic: "[{\"name\": \"eth0\", \"ip\": \"10.0.0.3\", \"subnet\": \"net0\"}]"
+spec:
+  containers:
+  - name: arktospod
+    image: localhost:5000/testpod
+    ports:
+      - containerPort: 443
```
zasherif, line 14: new line

mizar/dp/mizar/operators/endpoints/endpoints_operator.py
```diff
@@ -269,3 +292,7 @@ def init_simple_endpoint_interfaces(self, worker_ip, spec):
         # The Interface service will create the veth peers for the interface and
         # allocate the mac addresses for us.
         return InterfaceServiceClient(worker_ip).InitializeInterfaces(interfaces)
+    
+    def delete_simple_endpoint(self, ep):
+        logger.info("Delete endpoint object assicated with interface {}".format(ep.name))
+        ep.delete_obj()
```
zasherif, line 298: new line

mizar/daemon/interface_service.py
```diff
@@ -219,8 +221,51 @@ def ConsumeInterfaces(self, request, context):
...
+    def _UpdateSubstrate(self, obj_name, droplet_ip, droplet_mac):
+        if obj_name not in self.known_substrates.keys():
+            self.known_substrates[obj_name] = droplet_ip
+        self.rpc.update_substrate_ep(droplet_ip, droplet_mac)
+
+    def _DeleteSubstrate(self, obj_name, droplet_ip, droplet_mac):
+        if obj_name in self.known_substrates.keys():
+            self.known_substrates.pop(obj_name)
+        self.rpc.update_substrate_ep(droplet_ip, droplet_mac)
```
zasherif, line 241: Can you have a comment here explaining this?

zasherif (changes requested):

mizar/daemon/interface_service.py
phudtran: We don't need this, it can be removed.
The transit xdp program does not need to update/delete the substrate information about its own node.

mizar/daemon/interface_service.py
```diff
@@ -371,3 +416,10 @@ def update_agent_metadata(self, interface):
         returncode, text = run_cmd(cmd)
         logger.info(
             "update_agent_metadata returns {} {}".format(returncode, text))
+
+    def update_substrate_ep(self, ip, mac):
```
phudtran, line 420: We don't need this, it can be removed.
The transit xdp program does not need to update/delete the substrate information about its own node.

mizar/daemon/interface_service.py
```diff
@@ -165,6 +166,8 @@ def _ProvisionVethInterface(self, interface, cni_params):
         # Configure the Transit Agent
         self._ConfigureTransitAgent(interface)
 
+        self._UpdateSubstrateInterface(interface)
```
phudtran, line 169: We don't need this, it can be removed.
The transit xdp program does not need to update/delete the substrate information about its own node.

mizar/obj/endpoint.py
```diff
@@ -111,6 +113,7 @@ def set_obj_spec(self, spec):
         self.status = get_spec_val('status', spec)
         self.vpc = get_spec_val('vpc', spec)
         self.net = get_spec_val('net', spec)
+        self.k8s_pod_name = get_spec_val('k8s_pod_name', spec)
```
phudtran, line 116: We don't have this field defined in the endpoint CRD.

mizar/daemon/interface_service.py
```diff
@@ -219,8 +221,51 @@ def ConsumeInterfaces(self, request, context):
...
+
+    def _DeleteVethInterface(self, interface):         
+        """
+        Delete a veth interface
+        """
+        veth_peer_index = get_iface_index(interface.veth.peer, self.iproute)
+        
+        self.iproute.link('del', index=veth_peer_index)
+
+        self._UpdateSubstrateInterface(interface, False)
```
phudtran, line 251: We don't need this, it can be removed.
The transit xdp program does not need to update/delete the substrate information about its own node.

mizar/daemon/interface_service.py
phudtran: We need to make sure the transit agent is unloaded from the interface before deleting it.

mizar/daemon/interface_service.py
```diff
@@ -33,6 +33,7 @@ def __init__(self):
         self.interfaces_q = queue.Queue()  # Used for Produce/Consume sync
         self.iproute = IPRoute()
         self.interfaces = {}  # In-memory tracking for Pod interfaces
+        self.known_substrates = {}
```
phudtran, line 36: We don't need this, it can be removed.
The transit xdp program does not need to update/delete the substrate information about its own node.

mizar/dp/mizar/operators/endpoints/endpoints_operator.py
```diff
@@ -202,15 +220,13 @@ def produce_simple_endpoint_interface(self, ep):
...
-        # At this point Mizar has provisioned the network
-        # TODO (cathy): mark the pod network as ready! Shall it be here or in
-        # the pod builtin wf.
         logger.info("Produced {}".format(interfaces))
 
     def create_simple_endpoints(self, interfaces, spec):
         """
         Create a simple endpoint object (calling the API operator)
         """
+        count = 0
```
phudtran, line 229: Why do we need this count?

phudtran (changes requested):
I left some comments regarding the delete RPCs.
The only RPC we need to keep track of in the interface_service(daemon) is the update_ep RPC not the update_substrate rpc. We also need to aggregate the update_ep calls between the operator and the daemon in one place and make delete_ep calls accordingly.

I think it is going to be very difficult to verify the correctness of the delete logic without porting over the rest of the tests from our old framework. I think we should defer the reference counting logic until we have the rests of the test.

mizar/store/operator_store.py
clu2xlu: This has to do with arktos renaming the pod. I notice that in arktos, say if we create a pod called pod1, arktos will append namespace and tenant name to the pod name. If nothing is specified, it will reassign the name follow by random hash.
I agree that this part of code needs modifications.

clu2xlu:
Tracking in PR #200

PR #186, opened 2020-07-13
Adding Arktos network annotations

clu2xlu:


mizar/dp/mizar/workflows/builtins/pods/create.py
```diff
@@ -61,25 +61,28 @@ def run(self):
...
-            'vpc': OBJ_DEFAULTS.default_ep_vpc,
-            'net': OBJ_DEFAULTS.default_ep_net,
+            'namespace': self.param.body['metadata'].get('namespace', 'default'),
+            'tenant': self.param.body['metadata'].get('tenant', ''),
+            'vpc': self.param.body['metadata'].get('labels', {}).get(
+                    'arktos.futurewei.com/network', OBJ_DEFAULTS.default_ep_vpc),
+            'net': self.param.body['metadata'].get('annotations', {}).get(
+                    'arktos.futurewei.com/nic', {}).get('subnet', OBJ_DEFAULTS.default_ep_net),
+            'ip': self.param.body['metadata'].get('annotations', {}).get(
+                    'arktos.futurewei.com/nic', {}).get('ip', ''),
```
zasherif: I am not sure about the correctness of the IP address here. The network operator will be allocating it. If Arktos will provide an IP for POD as a mandatory field, there must be a way to tell the network operator not to allocate the IP address. Would you please ensure that?

mizar/dp/mizar/workflows/builtins/pods/create.py
```diff
@@ -61,25 +61,28 @@ def run(self):
...
 
         logger.info("Pod spec {}".format(spec))
         spec['vni'] = vpc_opr.store_get(spec['vpc']).vni
         spec['droplet'] = droplet_opr.store_get_by_ip(spec['hostIP'])
 
-        # TODO (cathy): make sure not to trigger init or create simple endpoint
+        # make sure not to trigger init or create simple endpoint
         # if Arktos network is already marked ready
-        if spec['phase'] != 'Pending':
+        if spec['phase'] != 'Pending' and spec['readiness'] == 'true':
```
zasherif: The readiness attribute is Arktos specific, I won't rely on that here. Also, one field the endpoint operator pass to the interface service is the computing provider. I believe we shall have an attribute here that reflects the provider. It can be a simple lookup for the an Arktos specific annotation to determine it's Arktos otherwise it is K8S. Would you please introduce that?

mizar/dp/mizar/workflows/builtins/pods/create.py
```diff
@@ -90,5 +93,6 @@ def run(self):
         # Create the corresponding simple endpoint objects
         endpoint_opr.create_simple_endpoints(interfaces, spec)
 
-        # TODO (cathy): in Arktos shall we mark the pod network ready here?
+        if 'arktos.futurewei.com/network' in self.param.body['metadata'].get('labels', {}):
+            endpoint_opr.annotate_builtin_pods(spec['name'], spec['namespace'])
```
zasherif, line 94: Have you tested it? Updating the readiness field is what allows Arktos to invoke the CNI. Ideally, We want this to happen after the operator produces the network interface.

zasherif (changes requested):
This is a good start, and we still have several iterations to get it right. The following stands out as major comments for me:
1. Is it correct that Arktos provides an IP address for the POD? If yes, there must be a way to prevent the network operator from allocating the IP address
1. Validate sequence of annotating the POD network readiness True.
1. Provide a field that specifies the compute cluster type (Arktos/K8s)

mizar/dp/mizar/operators/endpoints/endpoints_operator.py
```diff
@@ -202,15 +220,13 @@ def produce_simple_endpoint_interface(self, ep):
...
-        # At this point Mizar has provisioned the network
-        # TODO (cathy): mark the pod network as ready! Shall it be here or in
-        # the pod builtin wf.
         logger.info("Produced {}".format(interfaces))
 
     def create_simple_endpoints(self, interfaces, spec):
         """
         Create a simple endpoint object (calling the API operator)
         """
+        count = 0
```
zasherif, line 229: What is the use of this count?

mizar/dp/mizar/operators/endpoints/endpoints_operator.py
```diff
@@ -222,6 +238,8 @@ def create_simple_endpoints(self, interfaces, spec):
             ep.set_vni(spec['vni'])
             ep.set_vpc(spec['vpc'])
             ep.set_net(spec['net'])
+            ep.set_ip(spec['interfaces'][count].get('ip', ''))
+            count += 1
```
zasherif, line 242: ditto

mizar/dp/mizar/operators/nets/nets_operator.py
```diff
@@ -140,9 +140,10 @@ def on_endpoint_init(self, body, spec, **kwargs):
 
     def allocate_endpoint(self, ep):
         n = self.store.get_net(ep.net)
-        if ep.ip == "":
+        ip = ep.ip
+        if ep.ip == '':
             ip = n.allocate_ip()
-            ep.set_ip(ip)
+        ep.set_ip(ip)
```
zasherif, line 146: I am not sure why this change is necessary. It is the same thing. Right?

mizar/dp/mizar/workflows/builtins/pods/create.py
```diff
@@ -41,44 +42,42 @@ def requires(self):
...
 
         spec = {
             'hostIP': self.param.body['status']['hostIP'],
             'name': self.param.body['metadata']['name'],
-            'namespace': self.param.body['metadata']['namespace'],
-            'tenant': '',
-            # TODO (Cathy) in case of arktos
-            # get VPC and net information from annotation
-            'vpc': OBJ_DEFAULTS.default_ep_vpc,
+            'type': 'k8s',
```
zasherif, line 49: type refers to provider type k8s or Arktos. Right? Would you please change it to a more specific key (e.g. compute_provider)?

mizar/dp/mizar/workflows/builtins/pods/create.py
```diff
@@ -41,44 +42,42 @@ def requires(self):
...
-        # TODO (cathy): make sure not to trigger init or create simple endpoint
+        if 'arktos.futurewei.com/network' in self.param.body['metadata'].get('labels', {}):
+            spec['type'] = 'arktos'
+
+        if 'arktos.futurewei.com/nic' in self.param.body['metadata'].get('annotations', {}):
+            net_config = self.param.body['metadata']['annotations']['arktos.futurewei.com/nic']
+            configs = json.loads(net_config)
+            spec['interfaces'] = configs
+            for config in configs:
+                spec['net'] = config.get('subnet', OBJ_DEFAULTS.default_ep_net)
```
zasherif, line 73: I don't understand the use of this. If configs have multiple interfaces each with a separate network,  this loop overrides the network to be the last 'subnet' or default. I don't think that this is the expected behavior. Would you please explain?

zasherif (changes requested):
I have a few more comments. Would you please add a test case in the end-to-end tests for this?

mizar/dp/mizar/workflows/builtins/pods/create.py
clu2xlu: All the interfaces associated to a pod should be in the same subnet (correct me if I am wrong). Thus, we should only take one of the interface subnet that was passed in using yaml. 
I think the correct way should be: 
     if subnet in config: 
          spec['net'] = config['subnet']

mizar/dp/mizar/operators/nets/nets_operator.py
clu2xlu: if ip is given, ep.set_ip(ip) will not be reached. So it needs to be move outside of if statement

mizar/dp/mizar/operators/endpoints/endpoints_operator.py
clu2xlu: spec['interfaces'] is a list which includes all the interfaces we need to create, thurs as we loop through each interface, we also need loop through spec['interfaces']. That's why I added count to loop through spec['interfaces']

mizar/dp/mizar/operators/endpoints/endpoints_operator.py
clu2xlu: To loop through spec['interfaces'] list

clu2xlu:
> I have a few more comments. Would you please add a test case in the end-to-end tests for this?

I will combine my two PRs (CNI-delete and this one ) and provide end to end test

mizar/dp/mizar/operators/endpoints/endpoints_operator.py
zasherif: I must be missing something here. How does an ```interface``` maps to ```spec['interfaces'][count]```?

mizar/dp/mizar/operators/nets/nets_operator.py
zasherif: if ep.ip has a value (which is being copied in line 143), then it is already set. Right?

mizar/dp/mizar/workflows/builtins/pods/create.py
zasherif: No, a POD can be multi-homed. Two interfaces on the pod may belong to different networks (and even different VPCs).

zasherif (changes requested):

mizar/dp/mizar/operators/endpoints/endpoints_operator.py
clu2xlu: spec["interface"] is a list of dictionaries: [create.py](https://github.com/futurewei-cloud/mizar/blob/dev-next/mizar/dp/mizar/workflows/builtins/pods/create.py#L73)
When multiple interfaces are passed in, spec["interface"] = [{"name": "eth0", "ip": "10.10.1.12", "subnet": "net1"}, {"name": "eth1", "ip": "10.10.1.13", "subnet": "net1"}, {"name": "eth2", "ip": "10.10.1.14", "subnet": "net1"}]

clu2xlu:
Merge with issue #185. Close this one

Issue #182, opened 2020-07-08
Setting MTU on master node to 9000 causes TLS Handshake timeouts.


phudtran:
Solution: Enable MTU discovery on the nodes with the command below.
`sysctl -w net.ipv4.tcp_mtu_probing=2`
Issue to be closed pending PRs

zasherif:
This uncovers one more issue we need to fix in the Kind setup; we need also the MTU on docker-0 to be 9000 to simulate a real-network.

phudtran:
Fixed.

PR #189, opened 2020-07-18
Host Endpoint  and Default System Service Handling

phudtran:
Features and Improvements:

- Host Endpoint implementation #78
- Create scaled endpoints for default system services
- Create a phantom gateway endpoint for network #110
- Refactor Dockerfiles for quicker dev environment setup with kind-setup.sh

Bug Fixes: (Thanks @zasherif for help with these!)

- Phantom Gateway fast path bug 
- Issue where PodCreate is called before droplets are ready
- TLS Handshake issue running daemon on master node #182

### DIFF
--- a/etc/crds/endpoints.crd.yaml
+++ b/etc/crds/endpoints.crd.yaml
@@ -37,6 +37,14 @@ spec:
     shortNames:
       - ep
       - eps
+  validation:
+    openAPIV3Schema:
+      properties:
+        backends:
+          type: array
+          items:
+            type: string
+          minItems: 0
   additionalPrinterColumns:
     - name: Type
       type: string
@@ -128,4 +136,3 @@ spec:
       priority: 0
       JSONPath: .spec.cnidelay
       description: Time to setup endpoint on droplet
-
--- a/etc/crds/endpoints.crd.yaml
+++ b/etc/crds/endpoints.crd.yaml
@@ -37,14 +37,6 @@ spec:
     shortNames:
       - ep
       - eps
-  validation:
-    openAPIV3Schema:
-      properties:
-        backends:
-          type: array
-          items:
-            type: string
-          minItems: 0
   additionalPrinterColumns:
     - name: Type
       type: string
--- a/etc/docker/daemon.Dockerfile
+++ b/etc/docker/daemon.Dockerfile
@@ -19,7 +19,7 @@
 # TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR
 # THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
-FROM python:3.7
+FROM fwnetworking/python_base:latest
 COPY . /var/mizar/
 RUN pip3 install /var/mizar/
 CMD mizard
--- a/kind-setup.sh
+++ b/kind-setup.sh
@@ -65,6 +65,10 @@ timeout=120
 
 kind delete cluster
 docker network rm kind 2> /dev/null
+# All interfaces in the network have an MTU of 9000 to
+# simulate a real datacenter. Since all container traffic
+# goes through the docker bridge, we must ensure the bridge
+# interfaces also has the same MTU to prevent ip fragmentation.
 docker network create -d bridge \
   --subnet=172.18.0.0/16 \
   --gateway=172.18.0.1 \
--- a/kind-setup.sh
+++ b/kind-setup.sh
@@ -67,9 +67,8 @@ kind delete cluster
 docker network rm kind 2> /dev/null
 docker network create -d bridge \
   --subnet=172.18.0.0/16 \
-  --gateway=172.18.0.1
+  --gateway=172.18.0.1 \
   --opt com.docker.network.driver.mtu=9000 \
-  --ipv6 \
   kind
 
 if [[ "$USER" == "dev" ]]; then
--- a/kind-setup.sh
+++ b/kind-setup.sh
@@ -64,6 +64,13 @@ NODES=${2:-3}
 timeout=120
 
 kind delete cluster
+docker network rm kind 2> /dev/null
+docker network create -d bridge \
+  --subnet=172.18.0.0/16 \
+  --gateway=172.18.0.1
+  --opt com.docker.network.driver.mtu=9000 \
+  --ipv6 \
+  kind
 
 if [[ "$USER" == "dev" ]]; then
     DOCKER_ACC="localhost:5000"
--- a/mizar/common/common.py
+++ b/mizar/common/common.py
@@ -211,12 +211,18 @@ def kube_list_obj(obj_api, plurals, list_callback):
         list_callback(name, spec, plurals)
 
 
-def kube_get_service_spec(core_api, service_name, service_namespace, get_callback):
-    response = core_api.read_namespaced_service(
-        name=service_name,
-        namespace=service_namespace
-    )
-    get_callback(response.spec)
+def kube_get_endpoints(core_api, service_name, service_namespace):
+    response = None
+    try:
+        response = core_api.read_namespaced_endpoints(
+            name=service_name,
+            namespace=service_namespace)
+        return response
+    except:
+        logger.debug("Failed to get endpoints for service {} in namespace {}".format(
+            service_name, service_namespace))
+    finally:
+        return response
 
 
 def get_spec_val(key, spec, default=""):
--- a/mizar/common/common.py
+++ b/mizar/common/common.py
@@ -202,7 +202,8 @@ def kube_list_obj(obj_api, plurals, list_callback):
         version="v1",
         namespace="default",
         plural=plurals,
-        watch=False)
+        watch=False,
+        timeout_seconds=10)
     items = response['items']
     for v in items:
         name = v['metadata']['name']
--- a/mizar/common/common.py
+++ b/mizar/common/common.py
@@ -216,9 +216,6 @@ def kube_get_service_spec(core_api, service_name, service_namespace, get_callbac
         name=service_name,
         namespace=service_namespace
     )
-    logger.info(response)
-    logger.info(response.spec)
-    logger.info(response.spec.cluster_ip)
     get_callback(response.spec)
 
 
--- a/mizar/common/common.py
+++ b/mizar/common/common.py
@@ -210,6 +210,17 @@ def kube_list_obj(obj_api, plurals, list_callback):
         list_callback(name, spec, plurals)
 
 
+def kube_get_service_spec(core_api, service_name, service_namespace, get_callback):
+    response = core_api.read_namespaced_service(
+        name=service_name,
+        namespace=service_namespace
+    )
+    logger.info(response)
+    logger.info(response.spec)
+    logger.info(response.spec.cluster_ip)
+    get_callback(response.spec)
+
+
 def get_spec_val(key, spec, default=""):
     return default if key not in spec else spec[key]
 
--- a/mizar/common/common.py
+++ b/mizar/common/common.py
@@ -23,12 +23,14 @@ import subprocess
 import ctypes
 import logging
 import luigi
+import kopf
 import datetime
 import dateutil.parser
 from kubernetes import watch, client
 from ctypes.util import find_library
 from mizar.common.constants import *
 from pathlib import Path
+from luigi.execution_summary import LuigiStatusCode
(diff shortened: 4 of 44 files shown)


etc/crds/endpoints.crd.yaml
```diff
@@ -37,6 +37,14 @@ spec:
...
       - ep
       - eps
+  validation:
+    openAPIV3Schema:
+      properties:
+        backends:
+          type: array
+          items:
+            type: string
+          minItems: 0
```
zasherif: I don't feel that we need to keep track of the backends here! This object will be very large and It seems to me that it's redundant data to the k8s endpoints object.

etc/docker/daemon.Dockerfile
```diff
@@ -19,7 +19,7 @@
 # TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR
 # THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
-FROM python:3.7
+FROM fwnetworking/python_base:latest
```
zasherif, line 22: Thank you for taking care of that! It was a real pain when testing :)

kind-setup.sh
```diff
@@ -64,6 +64,12 @@ NODES=${2:-3}
 timeout=120
 
 kind delete cluster
+docker network rm kind 2> /dev/null
+docker network create -d bridge \
+  --subnet=172.18.0.0/16 \
+  --gateway=172.18.0.1 \
+  --opt com.docker.network.driver.mtu=9000 \
+  kind
```
zasherif, line 76: Would you please add a comment here explaining why we needed the MTU of the docker network to be 9000?

mizar/common/common.py
```diff
@@ -202,14 +202,29 @@ def kube_list_obj(obj_api, plurals, list_callback):
         version="v1",
         namespace="default",
         plural=plurals,
-        watch=False)
+        watch=False,
+        timeout_seconds=10)
```
zasherif: How does the caller handle a timeout? Did you notice that happening in tests?

mizar/daemon/interface_service.py
```diff
@@ -144,14 +144,14 @@ def _ConsumeInterfaces(self, pod_name, cni_params):
         logger.info("Consumed {}".format(interfaces))
         return interfaces
 
-    def _ProvisionInterface(self, interface, cni_params):
+    def _ProvisionInterface(self, interface, cni_params, update=True):
```
zasherif: What is the use of this update flag for a regular interface? I see that update is only False in case of activating host endpoint, I think it will be more readable to remove this Flag and have a separate ProvisionHostInterface function.

mizar/daemon/interface_service.py
```diff
@@ -144,14 +144,14 @@ def _ConsumeInterfaces(self, pod_name, cni_params):
...
         """
         Provision the interface according to its type
         """
         if interface.interface_type == InterfaceType.veth:
-            self._ProvisionVethInterface(interface, cni_params)
+            self._ProvisionVethInterface(interface, cni_params, update)
 
-    def _ProvisionVethInterface(self, interface, cni_params):
+    def _ProvisionVethInterface(self, interface, cni_params, update):
```
zasherif: ditto

mizar/daemon/interface_service.py
```diff
@@ -163,21 +163,22 @@ def _ProvisionVethInterface(self, interface, cni_params):
         self.iproute.link('set', index=veth_peer_index, state='up', mtu=9000)
 
         # Configure the Transit Agent
-        self._ConfigureTransitAgent(interface)
+        self._ConfigureTransitAgent(interface, update)
```
zasherif: ditto

mizar/daemon/interface_service.py
```diff
@@ -163,21 +163,22 @@ def _ProvisionVethInterface(self, interface, cni_params):
...
 
-    def _ConfigureTransitAgent(self, interface):
+    def _ConfigureTransitAgent(self, interface, update):
```
zasherif: ditto

mizar/daemon/interface_service.py
```diff
@@ -163,21 +163,22 @@ def _ProvisionVethInterface(self, interface, cni_params):
...
         """
         Load the Transit Agent XDP program, program all the bouncer substrate,
         update the agent metadata and endpoint.
         """
         self.rpc.load_transit_agent_xdp(interface)
 
-        for bouncer in interface.bouncers:
-            self.rpc.update_agent_substrate_ep(
-                interface.veth.peer, bouncer.ip_address, bouncer.mac)
+        if update:
```
zasherif: Why does this function skip the bouncer update, metadata, in case of host endpoint? How does it even work without these?

mizar/dp/mizar/operators/bouncers/bouncers_operator.py
```diff
@@ -88,7 +87,7 @@ def update_endpoint_with_bouncers(self, ep):
         for key in bouncers:
             bouncers[key].update_eps(eps)
 
-        if ep.type == OBJ_DEFAULTS.ep_type_simple:
+        if ep.droplet_obj:
```
zasherif: Skipping the update_bouncers_list if the droplet_obj is set seems confusing to me. When exactly this object is set is an implementation thing, which is likely to be decided in another part of the code. Making the condition based on the endpoint type is more concrete and readable. This is because it matches the intention and usage of the endpoint itself.

mizar/dp/mizar/operators/endpoints/endpoints_operator.py
```diff
@@ -94,9 +94,10 @@ def update_bouncer_with_endpoints(self, bouncer):
     def update_endpoints_with_bouncers(self, bouncer):
         eps = self.store.get_eps_in_net(bouncer.net).values()
         for ep in eps:
-            ep.update_bouncers(set([bouncer]))
+            if ep.droplet_obj:
+                ep.update_bouncers({"bouncer.name": bouncer})
```
zasherif: ditto, the condition should be on endpoint type.

mizar/dp/mizar/operators/endpoints/endpoints_operator.py
```diff
@@ -94,9 +94,10 @@ def update_bouncer_with_endpoints(self, bouncer):
...
 
-    def create_scaled_endpoint(self, name, spec):
+    def create_scaled_endpoint(self, name, spec, backends, namespace="default"):
```
zasherif: Why it is necessary to query the backends (endpoints) and passing it to create_scaled_endpoints? We have a built-in workflow for the K8S endpoints, which we use to track the backends of the scaled endpoints. Why this needs to be different for system services?

mizar/dp/mizar/operators/endpoints/endpoints_operator.py
```diff
@@ -109,15 +110,34 @@ def create_scaled_endpoint(self, name, spec):
...
+        backends = set()
+        response = kube_get_endpoints(self.core_api, name, namespace)
+        if response and response.subsets:
+            backends.add(response.subsets[0].addresses[0].ip)
+            ep.set_backends(list(backends))
+        logger.info(
+            "Added backends: {} to scaled endpoint {} ".format(backends, name))
         ep.create_obj()
-        self.annotate_builtin_endpoints(name)
+        self.annotate_builtin_endpoints(name, namespace)
```
zasherif, line 114: I remember we agreed to just remove the annotation filter, and create the scaled endpoints for all the services.

mizar/dp/mizar/workflows/builtins/pods/triggers.py
```diff
@@ -38,5 +38,11 @@ async def builtins_on_pod(body, spec, **kwargs):
...
-
-    run_workflow(wffactory().k8sPodCreate(param=param))
+    wf = wffactory().k8sPodCreate(param=param)
+    run_workflow(wf)
+    if wf.temporary_error:
+        raise kopf.TemporaryError(
+            "Temporary Error: {}".format(wf.error), delay=1)
+    if wf.permanent_error:
+        raise kopf.PermanentError(
+            "Permanent Error: {}".format(wf.error), delay=1)
```
zasherif: We need to repeat this pattern across all triggers. Best to implement this in run_workflow function so we don't copy this code again.

mizar/obj/bouncer.py
```diff
@@ -135,10 +135,17 @@ def load_transit_xdp_pipeline_stage(self):
...
+            if ep.droplet_obj:
                 self._update_simple_ep(ep)
             if ep.type == OBJ_DEFAULTS.ep_type_scaled:
                 self._update_scaled_ep(ep)
+            if ep.type == OBJ_DEFAULTS.ep_type_gateway:
+                self.update_gw_ep(ep)
+
+    def update_gw_ep(self, ep):
+        logger.info("Update gateway endpoint")
+        ep.set_backends([self.ip])
```
zasherif, line 147: The gw does not have a backend, why this statement is needed?

zasherif (changes requested):
Looks good in general. I have several major comments though. This is the first round of reviews, I will add more after we discuss these.

mizar/obj/bouncer.py
phudtran: We talked about setting the remote_ip of the gateway to be bouncer ip that the RPC is being called on.

mizar/dp/mizar/workflows/builtins/pods/triggers.py
phudtran: Ok will do!

mizar/dp/mizar/operators/endpoints/endpoints_operator.py
phudtran: I kept the annotation filters here because system endpoints without corresponding services such as the Kube-controller-manager, and Kube-scheduler get updated continuously (about every few seconds) which  then triggers our update workflow.

mizar/dp/mizar/operators/bouncers/bouncers_operator.py
phudtran: Ok, I will change this!

mizar/dp/mizar/operators/endpoints/endpoints_operator.py
phudtran: Will change this.

etc/crds/endpoints.crd.yaml
phudtran: I'll remove this.
This was originally needed since when system services come up, the bouncers are not yet provisioned.
Thus we had to rely on the bouncers to call the update RPCs. 
The kopf retry mechanism fixes this.

mizar/dp/mizar/operators/endpoints/endpoints_operator.py
phudtran: Thanks for catching this! This variable was left over from some old testing.

mizar/daemon/interface_service.py
phudtran: This works because the bouncers make the update when they come up.
I have this because the bouncers are not up upon host endpoint creation.
I can remove the flag and just move all the necessary calls to the ActiveHostInterface function

mizar/common/common.py
phudtran: I dug a bit deeper into the libraries' code and noticed that the default timeout for all requests is set to "None" or no timeout.
I originally added this hoping for a retry but that is not the case, this parameter will just timeout the request on the client side after a specified amount of time.
I've removed it for now, but do you think we should set a timeout for all of our requests?

mizar/obj/bouncer.py
zasherif: Yeah, I forgot that.

setup.py
```diff
@@ -9,17 +9,5 @@
...
-        'ipaddress',
-        'pyroute2',
-        'rpyc',
-        'kubernetes==11.0.0',
-        'luigi==2.8.12',
-        'grpcio',
-        'protobuf',
-        'fs'
-    ]
+    install_requires=[]
```
zasherif: I don't think we should remove these. If they are already installed in the image, that's fine. But we need to keep the dependencies in the setup script anyway.

zasherif (approved):
One minor comment on the setup script. Ow, lgtm

setup.py
phudtran: Ok I'll re-add them

PR #193, opened 2020-07-22
Fix issue: Failed to create bouncer. 'set' object has no attribute 'values'

Hong-Chang:
This is hot fix for issue that cannot create bouncer. Error message is 'set' object has no attribute 'values'


phudtran (approved):
lgtm

zasherif (changes requested):
Thanks for fixing this; I think the right fix is to consume this as a set here: https://github.com/futurewei-cloud/mizar/blob/dev-next/mizar/obj/endpoint.py#L237 (i.e. remove the .values() ). We should probably use type hints in the future since we are running all this in python3.

PR #195, opened 2020-07-22
Add monitoring service

ShixiongQi:
Build the monitoring service (kernel) in a separate function (trn_trace_prog_kern.c)


src/include/trn_datamodel.h
```diff
@@ -137,3 +137,16 @@ struct scaled_endpoint_remote_t {
...
+ */
+struct metrics_record {
+	__u64 n_pkts; /* # of pkts received */
+	__u64 total_bytes_rx; /* to calcualte rx bandwidth */
+	__u64 total_bytes_tx; /* to calcualte tx bandwidth */
+	__u64 n_tx; /* TX PPS for the Bouncer (which is symmetric) == RX PPS for the bouncer */
+	__u64 n_pass; /* forwarded to kernel */
+	__u64 n_drop; /* packets to be dropped*/
+	__u64 n_redirect; /* packets to be redirected*/
+} __attribute__((packed));
```
zasherif: new line

src/xdp/trn_trace_prog_kern.c
```diff
@@ -0,0 +1,77 @@
...
+#include "bpf_legacy.h"
+
+#include "trn_datamodel.h"
+#include "trn_transit_xdp_maps.h"
+#include "trn_kern.h"
+
+int _version SEC("version") = 1;
+
+SEC("trace_metrics_per_packet")
+int trace_metrics_per_packet(int *act, struct transit_packet *pkt)
```
zasherif: Do we need a section here for ebpf-to-ebpf call? Would you please change that to:
```
static __inline int trace_metrics_per_packet(int *act, struct transit_packet *pkt) {
```

src/xdp/trn_trace_prog_kern.c
```diff
@@ -0,0 +1,77 @@
...
+{
+	// struct metrics_record *rec;
```
zasherif: remove comment

src/xdp/trn_trace_prog_kern.c
```diff
@@ -0,0 +1,77 @@
...
+		bpf_debug("[Transit:%d:] ABORTED: No metrics table found\n",
+			  __LINE__);
+		return XDP_ABORTED;
+	}
+
+	pkt->rec->n_pkts++;
+	pkt->rec->total_bytes_rx += （pkt->data_end - pkt->data）* sizeof(long) / sizeof(void);
+
+	if (*act == XDP_TX)
+		*act = bpf_redirect_map(&interfaces_map, pkt->itf_idx, 0);
```
zasherif: We don't need this here (after changing to __inline__), you basically count the XDP actions and return "without modifying the action".

src/xdp/trn_trace_prog_kern.c
```diff
@@ -0,0 +1,77 @@
...
+	if (*act == XDP_PASS)
+		pkt->rec->n_pass++;
+	if (*act == XDP_DROP)
+		pkt->rec->n_drop++;
+	if (*act == XDP_TX) {
+		pkt->rec->n_tx++;
+		pkt->rec->total_bytes_tx += （pkt->data_end - pkt->data）* sizeof(long) / sizeof(void);
+	}
+	if (*act == XDP_ABORTED)
+		return XDP_ABORTED;
```
zasherif: would you please add n_aborted count as well

src/xdp/trn_trace_prog_kern.c
```diff
@@ -0,0 +1,77 @@
...
+	// struct metrics_record *rec;
+	__u32 key = 0;
+
+	/* Look up the entry in the metrics table */	
+	pkt->rec = bpf_map_lookup_elem(&metrics_table, &key);
+	if (!rec) {
+		bpf_debug("[Transit:%d:] ABORTED: No metrics table found\n",
+			  __LINE__);
+		return XDP_ABORTED;
+	}
```
zasherif: You can do this lookup before calling the function and abort there. This function call shall imply always success.

src/xdp/trn_trace_prog_kern.c
```diff
@@ -0,0 +1,77 @@
...
+	if (*act == XDP_ABORTED)
+		return XDP_ABORTED;
+	if (*act == XDP_REDIRECT)
+		pkt->rec->n_redirect++;
+
+// #define DEBUG_XDP_MONITORING
+#ifdef DEBUG_XDP_MONITORING
+	if (pkt->rec->n_pkts != (pkt->rec->n_tx + pkt->rec->n_pass + pkt->rec->n_drop + pkt->rec->redirect))
+		return XDP_ABORTED;
+#endif
```
zasherif: I am not sure what is that for? There is a DEBUG directive you may want to use that. Or let's have DEBUG_XDP_MONIROTING passed in the makefile. I'd just make it simple and reuse DEBUG. Also need not to return Aborted, you can use 
```bpf_debug();``` (only active during debugging) and observe the kernel traces.

**Review by zasherif — CHANGES_REQUESTED** [review]
Thanks for contributing this part. Great start. I have few comments on the XDP part, and still looking into the user space program.

File: src/xdp/trn_trace_prog_kern.c (hunk 2 above)
Comment (ShixiongQi) [reply src/xdp/trn_trace_prog_kern.c]: agreed

File: src/xdp/trn_trace_prog_kern.c [hunk 8]
```diff
@@ -0,0 +1,77 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+/**
+ * @file trn_trace_prog_kern.c
+ * @author Sherif Abdelwahab (@zasherif)
+ *         Phu Tran          (@phudtran)
```
zasherif: Please fix these
```
@author Shixiong Qi (@ShixiongQi )
```

src/xdp/trn_trace_prog_kern.c
ShixiongQi: Fixed.
And I also change the trn_trace_prog_kern.c to trn_trace_prog_kern.h and include it in trn_transit_xdp.c

src/xdp/trn_trace_prog_kern.c
ShixiongQi: Ok, I move the metrics table lookup into the __transit_ function in trn_transit_xdp.c, before calling the _trace_metrics_per_packet_

src/xdp/trn_transit_xdp_maps.h
```diff
@@ -34,6 +34,17 @@
...
+ * eBPF map for metrics collector
+ */
+struct bpf_map_def SEC("maps") metrics_table = {
+	.type = BPF_MAP_TYPE_PERCPU_ARRAY, /* Which types? PERCPU? PROG? */
+	.key_size = sizeof(__u32),
+	.value_size = sizeof(struct metrics_record),
+	.max_entries = 1,
+};
+BPF_ANNOTATE_KV_PAIR(metrics_table, __u32, struct metrics_record);
+
```
zasherif: Would you please move this definition to the end of the file. I think it might cause a map parsing issue (did not test it though). Let's also make sure that basic tests won't be impacted by your change here. We can help you with running these tests.

Also in the transit daemon, we should take care of maintaining this map info here: https://github.com/futurewei-cloud/mizar/blob/dev-next/src/dmn/trn_transit_xdp_usr.c#L90

src/xdp/trn_transit_xdp.c
```diff
@@ -784,14 +785,27 @@ int _transit(struct xdp_md *ctx)
...
 	int action = trn_process_eth(&pkt);
 
+	/* Look up the entry in the metrics table */
+	__u32 metrics_key = 0;	
+	pkt->rec = bpf_map_lookup_elem(&metrics_table, &metrics_key);
+
+	if (!pkt->rec) {
+		bpf_debug("[Transit:%d:] ABORTED: No metrics table found\n",
+			  __LINE__);
+		return XDP_ABORTED; // DO we need to count n_aborted here?
```
zasherif: I don't think it's possible to count n_aborted here because the metric struct is not even there. This is a fatal-error that nothing will work after it :)

src/xdp/trn_transit_xdp.c
```diff
@@ -784,14 +785,27 @@ int _transit(struct xdp_md *ctx)
 	if (!itf) {
 		bpf_debug("[Transit:%d:] ABORTED: Bad configuration\n",
 			  __LINE__);
-		return XDP_ABORTED;
+		return XDP_ABORTED; // DO we need to count n_aborted here?
```
zasherif: No count for n_aborted, see review comment below. I think you can remove this comment from the code.

src/xdp/trn_trace_prog_usr.c
```diff
@@ -0,0 +1,463 @@
...
+	struct bpf_map *map;
+	int interval = 2; /* sampling interval */
+	const char *bpf_map_name = "metrics_table";
+
+	map = (struct bpf_map*)malloc(sizeof(struct bpf_map));
+	map->name = bpf_map_name;
+	// where to find the file descriptor. It should belong to the bpf program running in transit xdp
+	// TODO: Find a way to get the fd, it is also needed in the stats_poll().
+	map->fd = 1;
+	map->def = NULL;
```
zasherif: You will basically need to iterate over the ebpf maps in the system and get the fd of the one matches the ```bpf_map_name```. You can do something similar to what the bpftool does.

src/xdp/trn_trace_prog_usr.c
```diff
@@ -0,0 +1,463 @@
...
+	name = bpf_map__name(map);
+	def  = bpf_map__def(map);
+	fd   = bpf_map__fd(map);
+
+	if (check_map(name, def, fd)) /* if map exists, return 0 */
+		return -1;
+
+	/* return 0 on successful maps check */
+	return 0;
+}
```
zasherif: Please see the review comments on how to obtain the ebpf map's FD.

src/xdp/trn_trace_prog_usr.c
```diff
@@ -0,0 +1,463 @@
...
+
+	res = clock_gettime(CLOCK_MONOTONIC, &t);
+	if (res < 0) {
+		fprintf(stderr, "Error with gettimeofday! (%i)\n", res);
+		exit(-1);
+	}
+	return (__u64) t.tv_sec * NANOSEC_PER_SEC + t.tv_nsec;
+}
+
+//TODO: Add n_aborted
```
zasherif: you still need this TODO?

src/xdp/trn_trace_prog_usr.c
```diff
@@ -0,0 +1,463 @@
...
+	if (err_only)
+		printf("\n%s\n", "???");
+
+	/* Trick to pretty printf with thousands separators use %' */
+	setlocale(LC_NUMERIC, "en_US");
+
+	while (1) {
+		swap(&prev, &rec);
+		stats_collect(map, rec);
+		stats_print(rec, prev, err_only);
```
zasherif: Does this work? you have the function commented out!

zasherif (changes requested):
I looked into the userspace part. Looks like still needs some work. See the issue of getting the maps FD by name, and let me know if you have questions about that.

ShixiongQi:
> I looked into the userspace part. Looks like still needs some work. See the issue of getting the maps FD by name, and let me know if you have questions about that.

Thanks Sherif, I'll have a look on it

src/xdp/trn_trace_prog_kern.h
```diff
@@ -0,0 +1,58 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+/**
+ * @file trn_trace_prog_kern.c
+ * @author ShixiongQi (@ShixiongQi)
+ *		   Sherif Abdelwahab (@zasherif)
```
zasherif: please remove my name. I did not author this code, and let's please make sure that the code is not copied from elsewhere and if so to have appropriate modifications, linking, or references. Thank you

src/xdp/trn_trace_prog_usr_get_map_fd.c
```diff
@@ -0,0 +1,368 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+/**
+ * @file trn_trace_prog_usr.c
+ * @author ShixiongQi (@ShixiongQi)
+ *		   Sherif Abdelwahab (@zasherif)
```
zasherif: ditto.

src/xdp/usr_prog_utility.h
```diff
@@ -0,0 +1,48 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+/**
+ * @file trn_trace_prog_usr.c
+ * @author ShixiongQi (@ShixiongQi)
+ *		   Sherif Abdelwahab (@zasherif)
```
zasherif: ditto

src/xdp/usr_prog_utility.c
```diff
@@ -0,0 +1,167 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+/**
+ * @file trn_trace_prog_usr.c
+ * @author ShixiongQi (@ShixiongQi)
+ *		   Sherif Abdelwahab (@zasherif)
```
zasherif: ditto

zasherif (changes requested):
Would you please verify the code is running in tests?

ShixiongQi:
I built some bpf maps in kernel and use the user monitoring program to test. Now the user program can read the metrics from the maps and calculate the stats. I'll move on to test it in the mizar environment

ShixiongQi:
I use iperf3 to generate some traffic in mizar network. The stats of the traffic are recorded by the monitoring program in the transit xdp and are stored in the BPF map called "metrics table".

I start a monitoring agent to poll the stats from the metrics table and print it out. The results are shown below:

checking map metrics_table

[Stats] Packets Per Second: 906.78
                                        TX Per Second: 0.00
                                        Pass Per Second: 6.00
                                        Drop Per Second: 0.00
                                        Redirect Per Second: 900.78
                                        Abort Per Second: 0.00

[Stats] Packets Per Second: 914.83
                                        TX Per Second: 0.00
                                        Pass Per Second: 1.00
                                        Drop Per Second: 0.00
                                        Redirect Per Second: 913.83
                                        Abort Per Second: 0.00

[Stats] Packets Per Second: 879.82
                                        TX Per Second: 0.00
                                        Pass Per Second: 6.00
                                        Drop Per Second: 0.00
                                        Redirect Per Second: 873.82
                                        Abort Per Second: 0.00

ShixiongQi:
Now the stats can be polled from the kernel with the monitoring agent (located at src/xdp/xdp_user_monitoring_prog).

The monitoring agent is now compiled with the library in XDP tutorial.

src/xdp/xdp_user_monitoring_prog/Makefile
```diff
@@ -0,0 +1,9 @@
+# SPDX-License-Identifier: (GPL-2.0 OR BSD-2-Clause)
```
phudtran: Can you change this file to module.mk so we can just compile everything from the root directory?

src/xdp/xdp_user_monitoring_prog/Makefile
```diff
@@ -0,0 +1,9 @@
...
+
+# Departing from the implicit _user.c scheme
+USER_TARGETS := xdp_user_monitoring_prog
+
+LIBBPF_DIR = ../../extern/libbpf/src/
+# COMMON_DIR = ../common/
+
+# include $(COMMON_DIR)/common.mk
```
phudtran: new line here

src/xdp/xdp_user_monitoring_prog/bpf_util.h
```diff
@@ -0,0 +1,61 @@
+/* SPDX-License-Identifier: GPL-2.0 */
+/* Copied from $(LINUX)/tools/testing/selftests/bpf/bpf_util.h */
```
phudtran: Since this file is a copy, we should move it to the src/extern/linux folder

phudtran (changes requested):
One minor suggestion. To make the userspace program cleaner, I think we can use a .h file for the constants, function prototypes, and change some of the smaller functions into macros.

src/xdp/xdp_user_monitoring_prog/Makefile
ShixiongQi: The xdp_user_monitoring_prog relies on the library support from the xdp_tutorial (https://github.com/xdp-project/xdp-tutorial). Curretnly, I compile the xdp_user_monitoring_prog under the directory of that. I'm not sure how to handle this, could you give me some suggestions?

ShixiongQi: Can we compile it under the directory of xdp_tutorial and then move it to "mizar/tools/"?

src/xdp/trn_transit_xdp.c
```diff
@@ -792,6 +793,22 @@ int _transit(struct xdp_md *ctx)
 
 	int action = trn_process_eth(&pkt);
 
+	/* 
+	 * Look up the entry in the metrics table, the key used is "0" 
+	 * This key need to keep constant with the key used in user prog.
+	 */
+	__u32 metrics_key = 0; 	
+	pkt.rec = bpf_map_lookup_elem(&metrics_table, &metrics_key);
```
ShixiongQi, line 802: @phudtran Hi Phu, which variable can be used as the unique id of a host here

src/xdp/trn_trace_prog_kern.h
```diff
@@ -0,0 +1,57 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ */
+
+#include <linux/bpf.h>
+//#include <bpf/bpf_helpers.h>
```
h-w-chen: do we need to keep comments here?

src/xdp/trn_trace_prog_kern.h
```diff
@@ -0,0 +1,57 @@
...
+
+// #include "bpf_legacy.h"
+
+#include "trn_datamodel.h"
+#include "trn_transit_xdp_maps.h"
+#include "trn_kern.h"
+
+//int _version SEC("version") = 1;
+
+static __inline void trace_metrics_per_packet(int *act, struct transit_packet *pkt)
```
h-w-chen, line 30: this function only uses pkt->rec; how about passing metrics_record pointer as arg instead of whole pkt.

src/xdp/trn_trace_prog_kern.h
```diff
@@ -0,0 +1,57 @@
...
+//#include <bpf/bpf_helpers.h>
+
+// #include "bpf_legacy.h"
+
+#include "trn_datamodel.h"
+#include "trn_transit_xdp_maps.h"
+#include "trn_kern.h"
+
+//int _version SEC("version") = 1;
+
```
h-w-chen: nit- what is the reason to pass int* act, instead of int act directly?

src/xdp/trn_trace_prog_kern.h
```diff
@@ -0,0 +1,57 @@
...
+	pkt->rec->n_pkts++;
+	pkt->rec->total_bytes_rx += (pkt->data_end - pkt->data)* sizeof(long) / sizeof(void);
+
+	if (*act == XDP_PASS)
+		pkt->rec->n_pass++;
+	if (*act == XDP_DROP)
+		pkt->rec->n_drop++;
+	if (*act == XDP_TX) {
+		pkt->rec->n_tx++;
+		pkt->rec->total_bytes_tx += (pkt->data_end - pkt->data) * sizeof(long) / sizeof(void);
```
h-w-chen, line 42: sizeof(long)/sizeof(void), in most arch, is 2. why does it need to times 2 to calculate the bytes? I may miss something, please help to clarify.

src/include/trn_datamodel.h
```diff
@@ -137,3 +137,17 @@ struct scaled_endpoint_remote_t {
...
+ */
+struct metrics_record {
+	__u64 n_pkts; /* # of pkts received */
+	__u64 total_bytes_rx; /* to calcualte rx bandwidth */
+	__u64 total_bytes_tx; /* to calcualte tx bandwidth */
+	__u64 n_tx; /* TX PPS for the Bouncer (which is symmetric) == RX PPS for the bouncer */
+	__u64 n_pass; /* forwarded to kernel */
+	__u64 n_drop; /* packets to be dropped*/
+	__u64 n_redirect; /* packets to be redirected*/
+	__u64 n_aborted;
```
h-w-chen, line 163: is n_aborted part of the metrics to record? if not, do we still keep it here?

src/xdp/trn_trace_prog_kern.h
```diff
@@ -0,0 +1,57 @@
...
+
+//int _version SEC("version") = 1;
+
+static __inline void trace_metrics_per_packet(int *act, struct transit_packet *pkt)
+{
+
+	pkt->rec->n_pkts++;
+	pkt->rec->total_bytes_rx += (pkt->data_end - pkt->data)* sizeof(long) / sizeof(void);
+
+	if (*act == XDP_PASS)
```
h-w-chen, line 36: nit - it seems a swicth/case srtcu fit more natrually.

src/xdp/trn_transit_xdp.c
```diff
@@ -792,6 +793,22 @@ int _transit(struct xdp_md *ctx)
 
 	int action = trn_process_eth(&pkt);
 
+	/* 
+	 * Look up the entry in the metrics table, the key used is "0"
```
h-w-chen, line 798: nit- since the metrics map is type array and only has 1 entry, the only valid key is 0.

src/xdp/xdp_user_monitoring_prog/Makefile
h-w-chen: I guess Phu required to rename this Makefile to module.mk, so that the mizar make can build this part in integrated/consistant way. I think it makes sense to have a consistant make system. How do you think?

If the build depends on xdp_tutorial library, one of the things we can do is moving the dependency lib to src/extern, just like libbpf lib. Another suggestion, is to remove user-space tool code from this PR, and figure it out later.

src/xdp/xdp_user_monitoring_prog/xdp_user_monitoring_prog.c
```diff
@@ -0,0 +1,560 @@
...
+#include <unistd.h>
+#include <sys/syscall.h>   /* For SYS_xxx definitions */
+#include <sys/ioctl.h>
+
+#include "bpf_util.h" /* bpf_num_possible_cpus */
+#include "trn_datamodel.h"
+
+#define EXIT_OK 		    0
+#define EXIT_FAIL		    1
+#define EXIT_FAIL_BPF		40
```
h-w-chen: super nit - to make this line in line with the above 2 (the alignment of values).

h-w-chen: is 40 arbitrary defined error code, or does it hav special reason?

src/xdp/xdp_user_monitoring_prog/xdp_user_monitoring_prog.c
```diff
@@ -0,0 +1,560 @@
...
+static int __check_map_fd_info(int map_fd, struct bpf_map_info *info,
+			       struct bpf_map_info *exp)
+{
+	/* Please be careful to "info_len". There will a possible error when
+	 * executing "bpf_obj_get_info_by_fd()":
+	 * 				Arguments list too long 
+	 */
+	struct bpf_map_info temp = {};
+	__u32 info_len = sizeof(temp);
+	// __u32 info_len = 40; //sizeof(*info);
```
h-w-chen: nit- please remove the comment purposed for temporary debugging only.

src/xdp/xdp_user_monitoring_prog/xdp_user_monitoring_prog.c
```diff
@@ -0,0 +1,560 @@
...
+};
+
+static int __check_map_fd_info(int map_fd, struct bpf_map_info *info,
+			       struct bpf_map_info *exp)
+{
+	/* Please be careful to "info_len". There will a possible error when
+	 * executing "bpf_obj_get_info_by_fd()":
+	 * 				Arguments list too long 
+	 */
+	struct bpf_map_info temp = {};
```
h-w-chen: it seems temp var is defined here merely to get the info_len value. Can we remove it?

src/xdp/xdp_user_monitoring_prog/xdp_user_monitoring_prog.c
```diff
@@ -0,0 +1,560 @@
...
+	struct metrics_record total;
+	struct metrics_record *cpu;
+};
+
+struct bpf_map {
+	char *name;
+	int fd;
+};
+
+static int __check_map_fd_info(int map_fd, struct bpf_map_info *info,
```
h-w-chen: the purpose of this function seems validing the map of map_fd has proper metadata of exp map. map info is the intermediary var, which seems not part of the signiture. Sugges to reconsider the necessity of the func, and necessity of *info param.

src/xdp/xdp_user_monitoring_prog/xdp_user_monitoring_prog.c
```diff
@@ -0,0 +1,560 @@
...
+		fprintf(stderr, "ERR: %s() "
+			"Map type(%d) mismatch expected type(%d)\n",
+			__func__, info->type, exp->type);
+		return EXIT_FAIL;
+	}
+
+	return 0;
+}
+
+static int __check_map(int map_fd, struct bpf_map_info *exp)
```
h-w-chen: this function seems shallow - does not do much. Can it just merge with  __check_map_fd_info body?

src/xdp/xdp_user_monitoring_prog/xdp_user_monitoring_prog.c
```diff
@@ -0,0 +1,560 @@
...
+
+	return __check_map_fd_info(map_fd, &info, exp);
+}
+
+static int check_map(const char *name, int fd)
+{
+	struct {
+		const char          *name;
+		struct bpf_map_info  info;
+	} maps[] = {
```
h-w-chen: is there a reason to define an array here? for future inclusion of more entries?

src/xdp/xdp_user_monitoring_prog/xdp_user_monitoring_prog.c
```diff
@@ -0,0 +1,560 @@
...
+}
+
+static int __check_map(int map_fd, struct bpf_map_info *exp)
+{
+	struct bpf_map_info info;
+
+	return __check_map_fd_info(map_fd, &info, exp);
+}
+
+static int check_map(const char *name, int fd)
```
h-w-chen: it seems this function only care about "metrics_table" map only. is there any reason to have general name as one param?

src/xdp/xdp_user_monitoring_prog/xdp_user_monitoring_prog.c
```diff
@@ -0,0 +1,560 @@
...
+
+	if (check_map(name, fd)) /* if map exists, return 0 */
+		return -1;
+
+	/* return 0 on successful maps check */
+	return 0;
+}
+
+#define NANOSEC_PER_SEC 1000000000 /* 10^9 */
+static __u64 gettime(void)
```
h-w-chen: nit - consider renaming get_time_in_nano_sec to reflect the precise purpose

src/xdp/xdp_user_monitoring_prog/xdp_user_monitoring_prog.c
```diff
@@ -0,0 +1,560 @@
...
+
+	res = clock_gettime(CLOCK_MONOTONIC, &t);
+	if (res < 0) {
+		fprintf(stderr, "Error with gettimeofday! (%i)\n", res);
+		exit(-1);
+	}
+	return (__u64) t.tv_sec * NANOSEC_PER_SEC + t.tv_nsec;
+}
+
+static bool map_collect_record(int fd, __u32 key, struct record *rec)
```
h-w-chen: is key always 0?

src/xdp/xdp_user_monitoring_prog/xdp_user_monitoring_prog.c
```diff
@@ -0,0 +1,560 @@
...
+		fprintf(stderr,
+			"ERR: bpf_map_lookup_elem failed key:0x%X\n", key);
+		return false;
+	}
+	/* Get time as close as possible to reading map contents */
+	rec->timestamp = gettime();
+
+	/* Record and sum values from each CPU */
+	for (i = 0; i < nr_cpus; i++) {
+		rec->cpu[i].n_pkts         =  values[i].n_pkts;
```
h-w-chen: do we really care about metric stats of individual cpu?

src/xdp/xdp_user_monitoring_prog/xdp_user_monitoring_prog.c
```diff
@@ -0,0 +1,560 @@
...
+}
+
+static double calc_period(struct record *r, struct record *p)
+{
+	double period_ = 0;
+	__u64 period = 0;
+
+	period = r->timestamp - p->timestamp;
+	if (period > 0)
+		period_ = ((double) period / NANOSEC_PER_SEC);
```
h-w-chen: if we only need is in term of sec, do we still need to keep timestamp in nano?

src/xdp/xdp_user_monitoring_prog/xdp_user_monitoring_prog.c
```diff
@@ -0,0 +1,560 @@
...
+					Pass Per Second: %'-10.2f\n \
+					Drop Per Second: %'-10.2f\n \
+					Redirect Per Second: %'-10.2f\n \
+					Abort Per Second: %'-10.2f\n\n\n",
+					pkts_per_second, tx_per_second, pass_per_second,
+					drop_per_second, redirect_per_second, abort_per_second);
+
+}
+
+static bool stats_collect(struct bpf_map *map, struct record *rec)
```
h-w-chen: function seems shallow. consider merge it with its implementaion body.

src/xdp/trn_kern.h
```diff
@@ -123,6 +123,9 @@ struct vlan_hdr {
 };
 
 struct transit_packet {
+	/* pointer to the metrics recorder */
+	struct metrics_record *rec;
```
h-w-chen, line 127: logically this metrics_record *rec not belong to packet. How about not putting ti as part of transit_packet?

h-w-chen (changes requested):
Thanks for the contribution and effort to add signifinant feature.

If the PR can be split into 2 or 3 ones, the first one about bpf prog/map changes, the remainders foucusing on user-space tools, that would make code review/merge process easier.

src/xdp/trn_transit_xdp.c
phudtran: I'm not aware of a pre-existing ID unique to each node that you can use here.
Perhaps you can push down the hostname of the node and use it as the key here.

src/xdp/trn_trace_prog_kern.h
ShixiongQi: fixed

src/xdp/trn_trace_prog_kern.h
ShixiongQi: This line is not needed, I removed it

src/xdp/trn_trace_prog_kern.h
ShixiongQi: My sense is to keep passing the pkt, as we need to use pkt->data and pkt->data_end to calculate bandwidth. Passing a ptr won't incur too much overhead

src/include/trn_datamodel.h
ShixiongQi: We can keep it for the moment, after we are clear about which metric is necessary, we can then decide to retain it  or not

src/xdp/xdp_user_monitoring_prog/xdp_user_monitoring_prog.c
ShixiongQi: I used the host IP instead of the key 0 in the latest  version. The key should be different for each host

src/xdp/xdp_user_monitoring_prog/xdp_user_monitoring_prog.c
ShixiongQi: Yes, we need to keep the stats for per cpu. This is to prevent the potential lock issue between different CPU cores

src/xdp/trn_transit_xdp.c
ShixiongQi: Thanks, it is fixed. I use host ip as the unique key per host

src/xdp/trn_transit_xdp.c
ShixiongQi: "kind" as a testing platform is tricky. The kind nodes will share the memory on the host machine, which means the eBPF map belong to each kind node co-locates on the same memory. So it would not be feasible to have the eBPF maps of different kind nodes use the same key. We cannot use the same key to differentiate the eBPF maps of different kind nodes.

Instead, I'll use the host ip as the unique key of each kind node, as each kind node has its unique IP in most case

src/xdp/xdp_user_monitoring_prog/xdp_user_monitoring_prog.c
ShixiongQi: agreed

vinaykul:
This PR is stale and superceded by my stats implementation XDP tail-call module based stats implementation. Closing.

Issue #187, opened 2020-07-14
Port Address Translation for Scaled Endpoint


phudtran:
Feature is needed for pods to communicated with API Server.
Example:
CoreDNS pods health check reply to API Server 10.96.0.1:443 -> 172.18.0.2:6443

PR #196, opened 2020-07-28
Scaled Endpoint Port Translation.

phudtran:
This PR adds support for port translation on the scaled endpoint #187.

### DIFF
--- a/mizar/common/constants.py
+++ b/mizar/common/constants.py
@@ -34,6 +34,8 @@ class CONSTANTS:
     ON_XDP_REDIRECT = "ON_XDP_REDIRECT"
     ON_XDP_DROP = "ON_XDP_DROP"
     ON_XDP_SCALED_EP = "ON_XDP_SCALED_EP"
+    IPPROTO_TCP = "6"
+    IPROTO_UDP = "17"
 
 
 class OBJ_STATUS:
@@ -91,6 +93,7 @@ class OBJ_DEFAULTS:
 
     mizar_service_annotation_key = "service.beta.kubernetes.io/mizar-scaled-endpoint-type"
     mizar_service_annotation_val = "scaled-endpoint"
+    kopf_annotation_key = "kopf.zalando.org/last-handled-configuration"
 
     kopf_max_retries = 5
 
--- a/mizar/common/constants.py
+++ b/mizar/common/constants.py
@@ -93,7 +93,6 @@ class OBJ_DEFAULTS:
 
     mizar_service_annotation_key = "service.beta.kubernetes.io/mizar-scaled-endpoint-type"
     mizar_service_annotation_val = "scaled-endpoint"
-    kopf_annotation_key = "kopf.zalando.org/last-handled-configuration"
 
     kopf_max_retries = 5
 
--- a/mizar/dp/mizar/operators/endpoints/endpoints_operator.py
+++ b/mizar/dp/mizar/operators/endpoints/endpoints_operator.py
@@ -166,7 +166,7 @@ class EndpointOperator(object):
                     backends.add(a['ip'])
             if "ports" in s:
                 for p in s['ports']:
-                    backend_ports.add(p['port'])
+                    backend_ports.add(str(p['port']))
         ep.set_backends(list(backends))
         ep.set_backend_ports(list(backend_ports))
         self.store_update(ep)
--- a/mizar/dp/mizar/operators/endpoints/endpoints_operator.py
+++ b/mizar/dp/mizar/operators/endpoints/endpoints_operator.py
@@ -111,12 +111,6 @@ class EndpointOperator(object):
         ep.set_mac(self.rand_mac())
         ep.set_type(OBJ_DEFAULTS.ep_type_scaled)
         ep.set_status(OBJ_STATUS.ep_status_init)
-        # logger.info("==SPEC")
-        # if "ports" in spec:
-        #     for port in spec["ports"]:
-        #         logger.info(port["protocol"])
-        #         logger.info(port["port"])
-        #         logger.info(port["targetPort"])
         ep.create_obj()
         self.annotate_builtin_endpoints(name, namespace)
 
--- a/mizar/dp/mizar/operators/endpoints/endpoints_operator.py
+++ b/mizar/dp/mizar/operators/endpoints/endpoints_operator.py
@@ -171,7 +171,7 @@ class EndpointOperator(object):
             return
         service_spec = list(service.metadata.annotations.values())
         json_spec = json.loads(service_spec[0])
-        # port = {frontend_port: [backend_port, protocol]}
+        # ports = {frontend_port: [backend_port, protocol]}
         if isinstance(json_spec, dict):
             for port in json_spec["spec"]["ports"]:
                 ports[port["port"]] = []
--- a/mizar/dp/mizar/operators/endpoints/endpoints_operator.py
+++ b/mizar/dp/mizar/operators/endpoints/endpoints_operator.py
@@ -21,6 +21,7 @@
 
 import logging
 import random
+import json
 from kubernetes import client, config
 from mizar.obj.endpoint import Endpoint
 from mizar.obj.bouncer import Bouncer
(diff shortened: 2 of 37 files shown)


mizar/dp/mizar/operators/endpoints/endpoints_operator.py
```diff
@@ -159,11 +159,16 @@ def update_scaled_endpoint_backend(self, name, spec):
         if ep is None:
             return None
         backends = set()
+        backend_ports = set()
```
zasherif: There is a problem with this. The backend_ports and backends may not have the same one-to-one mapping with sets. It can be ordered. It is better to associate a port with a backend in a separate class or simply a tuple. It's unlikely to happen, but it can happen.

src/dmn/trn_rpc_protocol_handlers_1.c
```diff
@@ -243,9 +243,14 @@ int *update_ep_1_svc(rpc_trn_endpoint_t *ep, struct svc_req *rqstp)
...
-	       epval.nremote_ips * sizeof(epval.remote_ips[0]));
-
+	if (epval.nremote_ips > 0) {
+		memcpy(epval.remote_ips, ep->remote_ips.remote_ips_val,
+		       epval.nremote_ips * sizeof(epval.remote_ips[0]));
+	}
+	if (epval.nremote_ports > 0) {
+		memcpy(epval.remote_ports, ep->remote_ports.remote_ports_val,
+		       epval.nremote_ports * sizeof(epval.remote_ports[0]));
+	}
```
zasherif: Shouldn't the nremote_ports always equal nremote_ips? Are you making a default value. If not, I believe these two should can be combined.

src/xdp/trn_transit_scaled_endpoint.c
```diff
@@ -393,7 +393,7 @@ static __inline int trn_scaled_ep_decide(struct transit_packet *pkt)
 		pkt->inner_ipv4_tuple.sport;
 
 	pkt->scaled_ep_opt->scaled_ep_data.target.dport =
-		pkt->inner_ipv4_tuple.dport;
+		bpf_htons(ep->remote_ports[remote_idx]);
```
zasherif: minor: You can save this conversion in the data-path by calling ```htons``` in the rpc implementation.

zasherif (changes requested):
a couple of minor comments. ow, lgtm

mizar/dp/mizar/operators/endpoints/endpoints_operator.py
phudtran: Thanks for the catch! I read up some more on Kubernetes services and it looks like we will need to do more work to support multi-port services.
https://kubernetes.io/docs/concepts/services-networking/service/#multi-port-services
In this case, I think we will instead need to have a one-to-one mapping between the frontend ports and backend ports.

This PR actually breaks the current scaled ep e2e tests since we run multiple servers on different ports (TCP, HTTP, UDP tests) in a single container.

src/dmn/trn_rpc_protocol_handlers_1.c
phudtran: Not all the time. We can have a single simple endpoint with no remote ports. Another scenario we need to support for Kubernetes is multi-port services where the number of backends can differ from the ports.
https://kubernetes.io/docs/concepts/services-networking/service/#multi-port-services

src/xdp/trn_transit_scaled_endpoint.c
phudtran: That makes sense, thanks for the catch!

phudtran:
Latest changes address #198 and #199

mizar/common/constants.py
```diff
@@ -91,6 +93,7 @@ class OBJ_DEFAULTS:
 
     mizar_service_annotation_key = "service.beta.kubernetes.io/mizar-scaled-endpoint-type"
     mizar_service_annotation_val = "scaled-endpoint"
+    kopf_annotation_key = "kopf.zalando.org/last-handled-configuration"
```
clu2xlu: Where is this key used

mizar/dp/mizar/operators/endpoints/endpoints_operator.py
```diff
@@ -164,6 +165,23 @@ def update_scaled_endpoint_backend(self, name, spec):
...
                     backends.add(a['ip'])
         ep.set_backends(list(backends))
+        ports = {}
+        service = kube_get_service(self.core_api, name, namespace)
+        if not service or not service.metadata or not service.metadata.annotations:
+            return
+        service_spec = list(service.metadata.annotations.values())
+        json_spec = json.loads(service_spec[0])
+        # port = {frontend_port: [backend_port, protocol]}
+        if isinstance(json_spec, dict):
```
clu2xlu, line 175: where is this "isinstance" func defined?

mizar/dp/mizar/workflows/builtins/pods/create.py
```diff
@@ -83,7 +85,7 @@ def run(self):
             return
         # Preexisting pods triggered when droplet objects are not yet created.
         if not spec['droplet']:
-            self.raise_temporary_error("Droplet not yet created")
+            self.raise_temporary_error("Droplet not yet created.")
```
clu2xlu, line 88: Instead of raising temporary error, can we consider using a while loop?

mizar/dp/mizar/workflows/builtins/services/create.py
```diff
@@ -53,8 +53,9 @@ def run(self):
         logger.info("Run {task}".format(task=self.__class__.__name__))
         if 'subsets' not in self.param.body:
             return
+        namespace = self.param.body["metadata"]["namespace"]
```
clu2xlu, line 56: use get to insert default value here in case there is no namespace passed.

mizar/dp/mizar/operators/endpoints/endpoints_operator.py
```diff
@@ -164,6 +165,23 @@ def update_scaled_endpoint_backend(self, name, spec):
                 for a in s['addresses']:
                     backends.add(a['ip'])
         ep.set_backends(list(backends))
+        ports = {}
+        service = kube_get_service(self.core_api, name, namespace)
+        if not service or not service.metadata or not service.metadata.annotations:
+            return
+        service_spec = list(service.metadata.annotations.values())
+        json_spec = json.loads(service_spec[0])
+        # port = {frontend_port: [backend_port, protocol]}
```
clu2xlu: do you mean ports, instead of port?

mizar/common/constants.py
phudtran: Thanks for the catch, this key is not used anymore.

mizar/dp/mizar/operators/endpoints/endpoints_operator.py
phudtran: It's defined by Python.

mizar/dp/mizar/workflows/builtins/pods/create.py
phudtran: No, a while loop will actually block the operator, and the condition will always be true.
Raising a temporary error here will make Kopf retry this operation after a given interval,
where the droplets are created.

mizar/dp/mizar/workflows/builtins/services/create.py
phudtran: This is from the Kubernetes spec when a service is created. We will use the default that is provided by Kubernetes, and not make up our own.

mizar/dp/mizar/operators/endpoints/endpoints_operator.py
phudtran: This is just a comment showing how the port dictionary maps the frontend port to a backend port and protocol.

clu2xlu: This was a bit confusing to me is because in the following lines,  you have: ports[port["port"]] = [] <-- this port dict is from spec
I think this dictionary maps you listed here is port(s) instead of port, right?

phudtran: Ah I see what you mean. Yeah, I'll change the name to make the correspondence more clear.

clu2xlu (approved):

PR #197, opened 2020-07-28
Convert vni from int to string. Fix test specs

clu2xlu:
1. VPC vni value needs to be stored as string 
2. Fixing test yam file's spec value.


zasherif (approved):

phudtran (approved):
lgtm

PR #223, opened 2020-08-28
gateway_design

zhangml:


liangbin-pub:
Suggest giving gateway project its own branch based from dev-next to house gateway specific document and code. Later on we can take it to its own repository.
Also, since we will put together a consolidated design based on both this and the DFT doc, we don't have to spent too much time on this one and should get it committed quickly.

liangbin-pub (changes requested):
Please change the target branch to Zeta and label as document

PR #224, opened 2020-08-28
Arktos pod create

clu2xlu:
Arktos pod create yaml file will include custom annotations and label to specify network information of the the pod, including arktos network, subnet, ip.


mizar/obj/endpoint.py
```diff
@@ -177,6 +178,9 @@ def set_vpc(self, vpc):
     def set_net(self, net):
         self.net = net
 
+    def set_k8s_pod_name(self, k8s_pod_name):
```
XiaoningDing, line 181: Is this a temporal design? It should be the other way around:  pod nic status or annotation tracks the id of endpoint.  With current design we won't be able to support multi-nic pods, and also Mizar got coupled with specific computing clients.

mizar/dp/mizar/operators/endpoints/endpoints_operator.py
```diff
@@ -144,6 +144,24 @@ def annotate_builtin_endpoints(self, name, namespace='default'):
                     "Retry updating annotating endpoints {}".format(name))
                 get_body = True
 
+    def annotate_builtin_pods(self, name, namespace='default'):
+        get_body = True
+        while get_body:
```
XiaoningDing: (namespace, name) could be duplicated among multiple tenants for Arktos. We need to think about how to work correctly with multi tenants here.

mizar/dp/mizar/operators/endpoints/endpoints_operator.py
```diff
@@ -258,12 +274,21 @@ def create_simple_endpoints(self, interfaces, spec):
...
+            OR arktos needs check wheter the ip and subnet is valid
+            """
+            ep.set_net(net_info.get('subnet', spec['net']))
+            ep.set_ip(net_info.get('ip', ''))
 
             ep.set_mac(interface.address.mac)
             ep.set_veth_name(interface.veth.name)
             ep.set_veth_peer(interface.veth.peer)
             ep.set_droplet(spec['droplet'].name)
+            ep.set_k8s_pod_name(spec['name'])
```
XiaoningDing: Is this a temporal design and will be changed after 9/15?  I think it should be the other way around:  Arkos side tracks the endpoint id in pod nic status.  So that Mizar can be neutral and can support any computing client. With current design it will limit what computing clients Mizar can support. Also it requires more and more such fields when we support more computing clients :)

mizar/dp/mizar/operators/endpoints/endpoints_operator.py
```diff
@@ -320,10 +345,14 @@ def init_simple_endpoint_interfaces(self, worker_ip, spec):
...
 
+            pod_provider = PodProvider.K8S
+            if spec['type'] == 'arktos':
+                pod_provider = PodProvider.ARKTOS
+
             interfaces_list.append(Interface(
                 interface_id=interface_id,
                 interface_type=InterfaceType.veth,
-                pod_provider=PodProvider.K8S,
+                pod_provider=pod_provider,
```
XiaoningDing, line 336: I know this is not introduced by your PR, just be curious why we need pod_provider.

mizar/dp/mizar/operators/endpoints/endpoints_operator.py
clu2xlu: This annotation is used to inform arktos that the pod creation is ready. Do we still need this annotation tho? Hongwei mentioned that it's good to have field. We can track the pod creation status by "STATUS" field in pod description. 

https://github.com/clu2xlu/mizar/blob/arktos-dns-service-deploy/mizar/dp/mizar/workflows/builtins/pods/create.py#L100-L101

https://github.com/futurewei-cloud/mizar/blame/dev-next/docs/design/mp_arktos.md#L205

mizar/dp/mizar/operators/endpoints/endpoints_operator.py
clu2xlu: This is Arktos if the pod is created in artkos. The default is k8s

mizar/dp/mizar/operators/endpoints/endpoints_operator.py
clu2xlu: I need remove this..... sorry, I missed this part

mizar/dp/mizar/operators/endpoints/endpoints_operator.py
clu2xlu: This annotation is added after the pod is set to be Provisioned, which does not work in two clusters env. Even if we want add this annotation, we should do it in network controller, right?

mizar/dp/mizar/workflows/builtins/pods/create.py
```diff
@@ -41,45 +42,45 @@ def requires(self):
...
+                    'arktos.futurewei.com/network-readiness', ''),
             'interfaces': [{'name': 'eth0'}]
         }
 
         logger.info("Pod spec {}".format(spec))
         spec['vni'] = vpc_opr.store_get(spec['vpc']).vni
         spec['droplet'] = droplet_opr.store_get_by_ip(spec['hostIP'])
-        # TODO (cathy): make sure not to trigger init or create simple endpoint
+
+        if 'arktos.futurewei.com/network' in self.param.body['metadata'].get('labels', {}):
```
XiaoningDing: It's better to define constants for strings like "arktos.futurewei.com/network" or "arktos.futurewei.com/nic"

mizar/common/constants.py
```diff
@@ -105,6 +104,9 @@ class RESOURCES:
     bouncers = "bouncers"
     dividers = "dividers"
 
+class COMPUTE_PROVIDER:
+    k8s = "k8s"
```
XiaoningDing: Better to use full name "Kubernetes" here. It's more formal :)

mizar/dp/mizar/operators/endpoints/endpoints_operator.py
```diff
@@ -194,7 +194,8 @@ def delete_endpoints_from_bouncers(self, bouncer):
     def delete_bouncer_from_endpoints(self, bouncer):
         eps = self.store.get_eps_in_net(bouncer.net).values()
         for ep in eps:
-            ep.update_bouncers(set([bouncer]), False)
+            if ep.type == OBJ_DEFAULTS.ep_type_simple or ep.type == OBJ_DEFAULTS.ep_type_host:
```
XiaoningDing, line 197: How about other EP types, do they need to update bouncer? My understanding is that for scaled endpoint and gateway endpoint we need to update bouncers too. Did I miss something?

mizar/store/operator_store.py
```diff
@@ -128,6 +133,14 @@ def delete_ep(self, name):
         l = len(self.eps_net_store[ep.net])
         if l == 0:
             del self.eps_net_store[ep.net]
+
+        if ep.k8s_pod_name in self.eps_pod_store.keys():
+            if name not in self.eps_pod_store[ep.k8s_pod_name]:
+                return ep
+            self.eps_pod_store[ep.k8s_pod_name].pop(name)
+            l = len(self.eps_pod_store[ep.k8s_pod_name])
```
XiaoningDing: Didn't you delete the "k8s_pod_name" field from the endpoint in the latest commit? Not sure how does this part work.

mizar/dp/mizar/workflows/builtins/pods/create.py
```diff
@@ -20,6 +20,7 @@
 # THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
 import logging
+import json
```
XiaoningDing, line 23: Sorry for a very basic question: why do we need to change pod workflows here?  My understanding is for the 9/15 release Enju/Hong's network controller will call endpoint operator directly to create endpoint objects. In that case we don't need pods workflow anymore, right?

XiaoningDing: Get the answer from scrum meeting. It's for native kubernetes.  We will discuss the design after current release. Close the comment.

mizar/store/operator_store.py
clu2xlu: Yes, thanks for catching this.

mizar/dp/mizar/operators/endpoints/endpoints_operator.py
XiaoningDing: Resolve it based on the discussion with Phu and Cathy.

mizar/common/constants.py
XiaoningDing: Please change the value to Kubernetes as well. Thanks.

XiaoningDing (approved):

PR #258, opened 2020-10-23
Dev mizar c2c

ghost:
Dynamic Interface.

### DIFF
--- a/mizar/common/rpc.py
+++ b/mizar/common/rpc.py
@@ -22,13 +22,14 @@
 import logging
 import json
 import netifaces
+import os
 from mizar.common.common import run_cmd
 
 logger = logging.getLogger()
 
 
 class TrnRpc:
-    def __init__(self, ip, mac, itf=netifaces.interfaces()[1], benchmark=False):
+    def __init__(self, ip, mac, itf=os.getenv("INET"), benchmark=False):
         self.ip = ip
         self.mac = mac
         self.phy_itf = itf
--- a/mizar/common/rpc.py
+++ b/mizar/common/rpc.py
@@ -21,13 +21,14 @@
 
 import logging
 import json
+import netifaces
 from mizar.common.common import run_cmd
 
 logger = logging.getLogger()
 
 
 class TrnRpc:
-    def __init__(self, ip, mac, itf='eth0', benchmark=False):
+    def __init__(self, ip, mac, itf=netifaces.interfaces()[1], benchmark=False):
         self.ip = ip
         self.mac = mac
         self.phy_itf = itf
--- a/mizar/obj/droplet.py
+++ b/mizar/obj/droplet.py
@@ -22,6 +22,7 @@
 import logging
 import json
 import netifaces
+import os
 from mizar.common.rpc import TrnRpc
 from mizar.common.constants import *
 from mizar.common.common import *
@@ -38,7 +39,7 @@ class Droplet(object):
         self.store = opr_store
         self.ip = ""
         self.mac = ""
-        self.phy_itf = netifaces.interfaces()[1]
+        self.phy_itf = os.getenv("INET")
         self.status = OBJ_STATUS.droplet_status_init
         self.known_substrates = {}
         self.known_bouncers = {}
--- a/mizar/obj/droplet.py
+++ b/mizar/obj/droplet.py
@@ -21,6 +21,7 @@
 
 import logging
 import json
+import netifaces
 from mizar.common.rpc import TrnRpc
 from mizar.common.constants import *
 from mizar.common.common import *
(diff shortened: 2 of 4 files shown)


mizar/common/rpc.py
```diff
@@ -21,13 +21,14 @@
...
 import json
+import netifaces
 from mizar.common.common import run_cmd
 
 logger = logging.getLogger()
 
 
 class TrnRpc:
-    def __init__(self, ip, mac, itf='eth0', benchmark=False):
+    def __init__(self, ip, mac, itf=netifaces.interfaces()[1], benchmark=False):
```
clu2xlu: We shouldn't use second item in the netifaces, because it's not always the second interface is the VM's default network interface.

mizar/obj/droplet.py
```diff
@@ -37,7 +38,7 @@ def __init__(self, name, obj_api, opr_store, spec=None):
         self.store = opr_store
         self.ip = ""
         self.mac = ""
-        self.phy_itf = 'eth0'
+        self.phy_itf = netifaces.interfaces()[1]
```
clu2xlu: Same as above comment, second interface is not always the default network interface.

clu2xlu (commented):
Overall, we cannot rely netifaces array to determine the default VM default network interface. 
We need have user to manually enter the name of the network interface, and use that user input as the phy_itf.

select_interface.py
```diff
@@ -0,0 +1,20 @@
+import inquirer
+import netifaces
+import os
+questions = [
+  inquirer.List('Network Interface',
+                message="Select Interfaces ?",
+                choices= netifaces.interfaces(),
+            ),
+]
+answers = inquirer.prompt(questions)
```
clu2xlu, line 10: We cannot rely on prompting for user input. If SRE were to deploy 1000 server, s/he will not enter the answer one by one.

XiaoningDing: Agree. Can we use a configuration file or a command line parameter to pass in the desired interface name?  If that doesn't exist, then return the  default interface name "eth0".

mizar/common/rpc.py
```diff
@@ -21,13 +21,15 @@
...
+import netifaces
+import os
 from mizar.common.common import run_cmd
 
 logger = logging.getLogger()
 
 
 class TrnRpc:
-    def __init__(self, ip, mac, itf='eth0', benchmark=False):
+    def __init__(self, ip, mac, itf=os.getenv("INET"), benchmark=False):
```
clu2xlu, line 32: Let's set the default value as "eth0" here: 
itf=os.getenv("INET") if os.getenv("INET") is not None else "eth0"

mizar/obj/droplet.py
```diff
@@ -37,7 +39,7 @@ def __init__(self, name, obj_api, opr_store, spec=None):
         self.store = opr_store
         self.ip = ""
         self.mac = ""
-        self.phy_itf = 'eth0'
+        self.phy_itf = os.getenv("INET")
```
clu2xlu, line 42: Let's set the default value as "eth0" here:
self.phy_itf=os.getenv("INET") if os.getenv("INET") is not None else "eth0"

mizar/obj/droplet.py
```diff
@@ -21,6 +21,8 @@
 
 import logging
 import json
+import netifaces
```
clu2xlu, line 24: remove netifaces, we use environment variable as the only source for interface name for now

select_interface.py
```diff
@@ -0,0 +1,20 @@
+import inquirer
```
clu2xlu, line 1: Remove select_interface.py file. We will not be promoting for user input

setup-machine-arktos.sh
```diff
@@ -182,6 +182,8 @@ pip3 install ipaddress
 pip3 install netaddr
 pip3 install kopf
 pip3 install PyYAML
+pip3 install netifaces
+pip3 install inquirer
```
clu2xlu, line 186: Remove both netifaces and inquirer. We won't be needing netifaces and inquirer as we only using environment variable to set interface name

clu2xlu (changes requested):
Please rebase with dev-next again and address comments. Feel free to submit a new PR if preferred.

vinaykul:
@Click2Cloud-Faizan This PR looks stale. If there are no plans to work on resolving the issues, please close the PR.

vinaykul:
> @Click2Cloud-Faizan This PR looks stale. If there are no plans to work on resolving the issues, please close the PR.

@Click2Cloud-Faizan No response from you, and this PR has been stale for several months now.
I assume that you don't plan on addressing the issues that have been called out, so I'm closing this stale PR.

PR #287, opened 2020-11-25
define bpf map for connection tracking

clu2xlu:
Connection tracking will be used to allow replies of an already established TCP or UDP connection. 
This PR is the initial step in defining connection tracking bpf map.

### DIFF
--- a/src/include/trn_datamodel.h
+++ b/src/include/trn_datamodel.h
@@ -167,3 +167,8 @@ struct vsip_ppo_t {
 	__u8 proto;
 	__be16 port;
 } __attribute__((packed));
+
+struct ipv4_ct_tuple_t {
+	struct vpc_key_t vpc;
+	struct ipv4_tuple_t tuple;
+} __attribute__((packed));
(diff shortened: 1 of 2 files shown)


src/include/trn_datamodel.h
```diff
@@ -167,3 +167,8 @@ struct vsip_ppo_t {
 	__u8 proto;
 	__be16 port;
 } __attribute__((packed));
+
+struct ipv4_ct_tuple_t {
+	struct vpc_key_t vpc;
```
h-w-chen, line 172: it seems the only info we care here is tunnel_id. Do you think which one make code easier to read: the simple use of __u64 tunnel_id, or wrapping with vpc_key_t?

src/include/trn_datamodel.h
```diff
@@ -167,3 +167,8 @@ struct vsip_ppo_t {
...
+	struct ipv4_tuple_t tuple;
```
h-w-chen, line 173: nic: suggest to use session or connection; tuple is generic term.

h-w-chen (approved):
generally it is OK. approved with a couple minor feedback.

PR #290, opened 2020-12-01
Add transit network policy map update

clu2xlu:
This partially resolves issue #271 #270 

This PR gets cli command input and trigger RPC call, then updates BPF map for network policies on the transit side.


src/cli/trn_cli_common.c
```diff
@@ -617,3 +617,64 @@ int trn_cli_parse_ebpf_prog_stage(const cJSON *jsonobj,
 	}
 	return 0;
 }
+
+int trn_cli_parse_network_policy_cidr(const cJSON *jsonobj,
+				      struct rpc_trn_vsip_cidr_t *cidrval)
+{
+	cJSON *prefixlen = cJSON_GetObjectItem(jsonobj, "prefixlen");
```
h-w-chen: prefixlen field seems better to be derived based on cidr subnet part. Suggest to consider changing the protocol to include only tunnel-id, local-ip, cidr-ip, cidr-subnet-prefixlen.

src/cli/trn_cli_common.c
```diff
@@ -617,3 +617,64 @@ int trn_cli_parse_ebpf_prog_stage(const cJSON *jsonobj,
...
+	if (local_ip != NULL && cJSON_IsString(local_ip)) {
+		struct sockaddr_in sa;
+		inet_pton(AF_INET, local_ip->valuestring, &(sa.sin_addr));
+		cidrval->local_ip = sa.sin_addr.s_addr;
+	} else {
+		print_err("Error: Network policy local IP is missing or non-string\n");
+		return -EINVAL;
+	}
+
+	if (remote_ip != NULL && cJSON_IsString(remote_ip)) {
```
h-w-chen: this block is very similar to lines [647..650]

src/cli/trn_cli_network_policy.c
```diff
@@ -0,0 +1,84 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+/**
+ * @file trn_cli_net.c
+ * @author Catherine Lu (@clu2)
+ *         
+ * @brief CLI subcommands related to networks
+ *
+ * @copyright Copyright (c) 2019 The Authors.
```
h-w-chen: nit: 2020?

src/dmn/trn_rpc_protocol_handlers_1.c
```diff
@@ -1257,3 +1259,56 @@ int *unload_transit_xdp_pipeline_stage_1_svc(rpc_trn_ebpf_prog_stage_t *argp,
...
+	char *itf = policy->interface;
+	struct vsip_cidr_t cidr;
+
+	TRN_LOG_INFO("update_transit_network_policy_1_svc service");
+
+	struct user_metadata_t *md = trn_itf_table_find(itf);
+	if (!md) {
+		TRN_LOG_ERROR("Cannot find interface metadata for %s", itf);
+		result = RPC_TRN_ERROR;
+		goto error;
```
h-w-chen, line 1279: why not return RPC_TRN_ERROR directly here?

src/dmn/trn_rpc_protocol_handlers_1.c
```diff
@@ -1257,3 +1259,56 @@ int *unload_transit_xdp_pipeline_stage_1_svc(rpc_trn_ebpf_prog_stage_t *argp,
...
+
+	cidr.tunnel_id = policy->tun_id;
+	cidr.prefixlen = policy->prefixlen;
+	cidr.local_ip = policy->local_ip;
+	cidr.remote_ip = policy->remote_ip;
+	__u64 bitmap = policy->bit_val;
+
+	if (policy->type == SUPPLEMENTARY) {
+		rc = trn_update_transit_supp_map(md, &cidr, bitmap);
+		if (rc != 0) {
```
h-w-chen: it seems trn_update_transit_supp_map(...) already have logging. Is there a reason we have another log entry here?

h-w-chen (changes requested):
please change the double-logging code; also, consider deriving prefixlen instead of relying on operator to calculate the final value.

src/cli/trn_cli_common.c
clu2xlu: Do you mean just change the naming from "prefixlen" to "cidr-subnet-prefixlen"?

h-w-chen: sort-of; what I suggested is to calculate the final prefixlen in transit daemon code, instead of relying on operator code. The rational is operator should not know about the detail; it knows only the destination CIDR (in your words cidr-subnet-prefixlen).

clu2xlu:
Move to PR #293

PR #293, opened 2020-12-02
Daemon update ingress network policy

clu2xlu:
This partially resolves issue #271 #270

This PR gets cli command input and trigger RPC call, then updates BPF map for network policies on the transit side.

### DIFF
--- a/src/cli/test/test_cli.c
+++ b/src/cli/test/test_cli.c
@@ -1977,7 +1977,7 @@ static void test_trn_cli_update_transit_network_policy_subcmd(void **state)
 		.interface = itf,
 		.tunid = 3,
 		.local_ip = 0x300000a,
-		.cidr_prefixlen = 112,
+		.cidr_prefixlen = 16,
 		.cidr_ip = 0x90000ac,
 		.cidr_type = 1,
 		.bit_val = 10
--- a/src/cli/test/test_cli.c
+++ b/src/cli/test/test_cli.c
@@ -494,11 +494,7 @@ static int check_policy_equal(const LargestIntegralType value,
 		return false;
 	}
 
-	if (policy->prefixlen != c_policy->prefixlen) {
-		return false;
-	}
-
-	if (policy->tun_id != c_policy->tun_id) {
+	if (policy->tunid != c_policy->tunid) {
 		return false;
 	}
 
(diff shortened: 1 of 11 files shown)


src/dmn/trn_transit_xdp_usr.c
```diff
@@ -579,3 +579,40 @@ uint32_t trn_get_interface_ipv4(int itf_idx)
...
+	int err = bpf_map_update_elem(md->ing_vsip_prim_map_fd, ipcidr, &bitmap, 0);
+	if (err) {
+		TRN_LOG_ERROR("Store Primary ingress map failed (err:%d).",
+			      err);
+		return 1;
+	}
+	return 0;
+}
+
+int trn_update_transit_supp_map(struct user_metadata_t *md, struct vsip_cidr_t *ipcidr,
```
XiaoningDing: Consider renaming these three functions to include "policy" term? Because there could be many maps, it would be confusing to understand the purpose of a function if it's only named "major_map".  Also "supp" seems not a good abbreviation :)

XiaoningDing (approved):

src/cli/test/test_cli.c
```diff
@@ -1890,6 +1937,92 @@ static void test_trn_cli_delete_agent_md_subcmd(void **state)
 	assert_int_equal(rc, -EINVAL);
 }
 
+static void test_trn_cli_update_transit_network_policy_subcmd(void **state)
```
h-w-chen, line 1940: kudos to have unit test code

src/cli/trn_cli_common.c
```diff
@@ -617,3 +617,68 @@ int trn_cli_parse_ebpf_prog_stage(const cJSON *jsonobj,
...
+	}
+
+	if (cidr_ip != NULL && cJSON_IsString(cidr_ip)) {
+		cidrval->cidr_ip = parse_ip_address(cidr_ip);
+	} else {
+		print_err("Error: Network policy CIDR IP is missing or non-string\n");
+		return -EINVAL;
+	}
+
+	// adding 96 = 0x01100000 offsets the missing bit
```
h-w-chen: 1. please explain the magic number 96 a bit.
2. what about putting the addition to daemon, instead of here at client?

h-w-chen (approved):
approved. left with a couple of minor comments for consideration.

PR #309, opened 2020-12-15
Added code for eth0 interface

click2cloud-akshay:
we had change the code for Interface now it able to pick the default interface .


click2cloud-akshay:
Need To add some Change .

PR #310, opened 2020-12-15
update network policy enforcement transit map

clu2xlu:

User inputs cli update-network-policy-enforcement-ingress to update entries in network policy enforcement bpf maps.
If Enforcement = 1, means pod / ep is enforced by network policy

The Cli command triggers RPC call and then perform bpf map update in transit agent.


h-w-chen (approved):
LGTM. 
kudos for having unit test cases.

PR #313, opened 2020-12-16
Delete transit network policy enforcement map

clu2xlu:

User inputs cli delete-network-policy-enforcement-ingress to delete entries in network policy enforcement bpf maps.

The Cli command triggers RPC call and then perform bpf map delete in transit agent.


h-w-chen (approved):
lgtm

PR #321, opened 2020-12-21
exposes maps of network policies and conn_track to ingress/egress xdp progs

### LINKED ISSUE #304: ensures egress xdp prog able to access ingress policy maps, and vise versa
This cross-referrence access is required by the new UDP policy check design taking conn_track into its consideration.
See mizar network policy design doc for details.

h-w-chen:
It fixes #304.

According to conn_track design of mizar network policy, in order to block the reply UDP packets after the policy updates and intends to block the originated udp traffic, policy enforcement logic needs signifiacnt enhancement. Due to udp being one directional protocol in nature, the _logical_ reply packets could go without the _loogical_ originated packet; this imposes some difficulty for policy neforcement, as there might be no packet emitted on the _originated direction_ wire, and the policy enforcement at the other direction has to handle w/o the help from the _original direction_. The proposal in the design is, for udp reply packet of a tracked connection, it consults the relevant policies to decide whether the _originated_ packet should be allowed or denied; based the this decision, the reply packet will be allowed or denied, respectively.

This PR is part of the implementation of the proposal: it exposes ingress policies to egress enforcement prog, and vise versa. The change of xdp prog to consulte polices and decide whether the udp reply packet should be allowed or denied is in the PR coming soon.

In the future implementation of conn_track, when tracked connections have states to differentaie being allowed or denied explicitly, the enforcement prog will disregard the state but relies on policy calculation when deciding where the reply packet should go or block.

### DIFF
--- a/src/dmn/trn_agent_xdp_usr.c
+++ b/src/dmn/trn_agent_xdp_usr.c
@@ -305,6 +305,11 @@ int trn_agent_bpf_maps_init(struct agent_user_metadata_t *md)
 	md->eg_vsip_ppo_map_fd		= bpf_map__fd(md->eg_vsip_ppo_map);
 	md->eg_vsip_supp_map_fd		= bpf_map__fd(md->eg_vsip_supp_map);
 	md->eg_vsip_except_map_fd	= bpf_map__fd(md->eg_vsip_except_map);
+	md->ing_vsip_enforce_map_fd	= bpf_map__fd(md->ing_vsip_enforce_map);
+	md->ing_vsip_prim_map_fd	= bpf_map__fd(md->ing_vsip_prim_map);
+	md->ing_vsip_ppo_map_fd		= bpf_map__fd(md->ing_vsip_ppo_map);
+	md->ing_vsip_supp_map_fd	= bpf_map__fd(md->ing_vsip_supp_map);
+	md->ing_vsip_except_map_fd	= bpf_map__fd(md->ing_vsip_except_map);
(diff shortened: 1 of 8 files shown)


src/dmn/trn_agent_xdp_usr.c
```diff
@@ -261,20 +266,28 @@ int trn_agent_bpf_maps_init(struct agent_user_metadata_t *md)
...
 	    !md->endpoints_map || !md->xdpcap_hook_map ||
 	    !md->fwd_flow_mod_cache_ref || !md->rev_flow_mod_cache_ref ||
 	    !md->ep_flow_host_cache_ref || !md->ep_host_cache_ref ||
 	    !md->eg_vsip_enforce_map || !md->eg_vsip_prim_map ||
 	    !md->eg_vsip_ppo_map || !md->eg_vsip_supp_map ||
-	    !md->eg_vsip_except_map || !md->conn_track_cache) {
+	    !md->eg_vsip_except_map ||  !md->ing_vsip_enforce_map ||
+	    !md->ing_vsip_prim_map || !md->ing_vsip_ppo_map ||
+	    !md->ing_vsip_supp_map || !md->ing_vsip_except_map ||
+	    !md->conn_track_cache) {
```
clu2xlu, line 290: Please also add fd

clu2xlu (changes requested):

src/dmn/trn_agent_xdp_usr.c
h-w-chen: will fix.

clu2xlu (approved):

PR #322, opened 2020-12-21
refactor: egress policy enforce has param of tunnel id and ipv4 tuple

h-w-chen:
This PR is trival refactoring by itself. It makes the egress_policy_enforce method to have same signiture as ingress_policy_enforce, which will make the future PR easy to check originated directional policy at the reply packet path (which is on the way soon).


clu2xlu (approved):
LGTM

PR #323, opened 2020-12-22
transit network policy protocol port map delete

clu2xlu:

User inputs cli delete-network-policy-protocol-port-ingress to delete entries in network policy protocol port bpf maps.

The Cli command triggers RPC call and then perform bpf map delete in transit agent.


h-w-chen (approved):
lgtm. go for it.

PR #325, opened 2020-12-23
egress and ingress xdp consult the derived policy to check reply UDP packets

### LINKED ISSUE #305: conn_track: ensure udp packets blocked as soon as policy is changed to deny
UDP packet may need to look up the policy rules even if the packet is "reply" based on conn_track data.

h-w-chen:
This closes #305.

Based on the mizar network policy and conn_track design, the decision to allow or deny UDP a reply packet should consult the _derived_ policy rule (which is derived from the one of originated direction), so that UDP stream in reply direction can be blocked after the originated policy has updated to deny the _originated_ traffic, the reply stream should go to halt as the reasonable intent. To check the _derived_ rule of a udp reply, ingress policy enforcement can simply work out the egress policy result of a synthesized egress packet, and egress policy enforcement does to ingress policy result of a synthesized ingress packet, respectively.

This PR moves all the policy check functions in the shared header which will be included by both ingress and egress xdp source code, and it implements the _derived_ policy check logic for ingress/egress policy reply packet check using hese functions.

There will be more PRs to introduce tracked connection states, which will cause slight change to the code that the PR makes.

### DIFF
--- a/src/xdp/trn_transit_xdp.c
+++ b/src/xdp/trn_transit_xdp.c
@@ -394,67 +394,6 @@ static __inline int trn_handle_scaled_ep_modify(struct transit_packet *pkt)
 	return XDP_TX;
 }
 
-/*
-   check if ingress network policy should be enforced to specific destination
-   return value:
-     0 (false) :    no ingress policy
-     non-0 (true) : need to enforce ingress policy check
-*/
-static __inline int is_ingress_enforced(__u64 tunnel_id, __be32 ip_addr)
-{
-	struct vsip_enforce_t vsip = { .tunnel_id = tunnel_id,
-				       .local_ip = ip_addr };
-	__u8 *v = bpf_map_lookup_elem(&ing_vsip_enforce_map, &vsip);
-	return v && *v;
-}
-
-/*
-   enforce_ingress_policy enforces egress network policy with the (incoming) packet
-   return value:
-     0: ingress policy allows this packet; no error
-    -1: ingress policy denies this packet; ingress policy denial error
-*/
-static __inline int enforce_ingress_policy(__u64 tunnel_id, const struct ipv4_tuple_t *ipv4_tuple)
-{
-	const __u32 full_vsip_cidr_prefix = (__u32)(sizeof(struct vsip_cidr_t) - sizeof(__u32)) * 8;
-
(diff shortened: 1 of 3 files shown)


src/xdp/trn_transit_xdp.c
```diff
@@ -509,8 +448,30 @@ static __inline int trn_process_inner_ip(struct transit_packet *pkt)
 
 	// todo: add conn_track related logic properly
 	if (pkt->inner_ipv4_tuple.protocol == IPPROTO_TCP || pkt->inner_ipv4_tuple.protocol == IPPROTO_UDP) {
-		if (conntrack_is_reply_of_tracked_conn(&conn_track_cache, tunnel_id, &pkt->inner_ipv4_tuple))
+		// todo: handle udp reply even when the originated connection is marked as denied;
```
XiaoningDing, line 451: "night" -> "might"?

src/xdp/trn_transit_xdp.c
```diff
@@ -509,8 +448,30 @@ static __inline int trn_process_inner_ip(struct transit_packet *pkt)
...
+					bpf_debug("[Transit:vpc 0x%lx] ABORTED: packet to 0x%x from 0x%x ingress denied, reply of a denied UDP conn\n",
+						tunnel_id,
+						bpf_ntohl(pkt->inner_ipv4_tuple.daddr),
+						bpf_ntohl(pkt->inner_ipv4_tuple.saddr));
+					// todo: to change conn state if it is available instead of removing it from the map
+					conntrack_remove_tcpudp_conn(&conn_track_cache, tunnel_id, &originated_tuple);
+					return XDP_ABORTED;
+				}
+			}
 			goto xdp_continue;
```
XiaoningDing, line 474: Can we avoid goto?  (can be done in later PRs, sure)

src/xdp/trn_transit_xdp.c
h-w-chen: typo; will fix.

src/xdp/trn_transit_xdp.c
h-w-chen: refactor is planned; will get rid of goto in next few PR soon.

clu2xlu (approved):

PR #327, opened 2020-12-24
switch to batch update for transit network policy cidr

### LINKED ISSUE #294: Change the policy update to batch style to improve performance

clu2xlu:
This partially resolves #294 

For network policy cidr updates, we were doing one by one bpf map update previously. To increase efficiency, we now switch to batch update

### DIFF
--- a/src/cli/test/test_cli.c
+++ b/src/cli/test/test_cli.c
@@ -2145,7 +2145,7 @@ static void test_trn_cli_update_transit_network_policy_subcmd(void **state)
 			  }]) };
 	char itf[] = "eth0";
 
-	struct rpc_trn_vsip_cidr_t exp_policy1 = {
+	struct rpc_trn_vsip_cidr_t policies[2] = {{
 		.interface = itf,
 		.tunid = 3,
 		.local_ip = 0x300000a,
@@ -2153,9 +2153,8 @@ static void test_trn_cli_update_transit_network_policy_subcmd(void **state)
 		.cidr_ip = 0x90000ac,
 		.cidr_type = 1,
 		.bit_val = 10
-	};
-
-	struct rpc_trn_vsip_cidr_t exp_policy2 = {
+	},
+	{
 		.interface = itf,
 		.tunid = 1,
 		.local_ip = 0x100000a,
@@ -2163,11 +2162,7 @@ static void test_trn_cli_update_transit_network_policy_subcmd(void **state)
 		.cidr_ip = 0x60000ac,
 		.cidr_type = 2,
 		.bit_val = 1
-	};
-
-	struct rpc_trn_vsip_cidr_t policies[2];
-	policies[0] = exp_policy1;
-	policies[1] = exp_policy2;
+	}};
 
 	/* Test call update_transit_network_policy successfully */
 	TEST_CASE("update-network-policy-ingress succeed with well formed policy json input");
--- a/src/cli/test/test_cli.c
+++ b/src/cli/test/test_cli.c
@@ -2110,35 +2110,42 @@ static void test_trn_cli_update_transit_network_policy_subcmd(void **state)
 	int update_transit_network_policy_1_ret_val = 0;
 
 	/* Test cases */
-	char *argv1[] = { "update-network-policy-ingress", "-i", "eth0", "-j", QUOTE({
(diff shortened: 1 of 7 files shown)


src/dmn/trn_transit_xdp_usr.c
```diff
@@ -639,39 +639,58 @@ uint32_t trn_get_interface_ipv4(int itf_idx)
...
 						  struct vsip_cidr_t *ipcidr,
-						  __u64 bitmap)
+						  __u64 *bitmap)
 {
-	int err = bpf_map_update_elem(md->ing_vsip_prim_map_fd, ipcidr, &bitmap, 0);
-	if (err) {
-		TRN_LOG_ERROR("Store Primary ingress map failed (err:%d).",
-			      err);
-		return 1;
+	int counter = (int)(sizeof(ipcidr) / sizeof(*ipcidr));
```
h-w-chen: whill this get desired value of element number? I am afraid not.

src/dmn/trn_transit_xdp_usr.h
```diff
@@ -214,15 +214,15 @@ int trn_remove_prog(struct user_metadata_t *md, unsigned int prog_idx);
 
 int trn_update_transit_network_policy_primary_map(struct user_metadata_t *md,
 						  struct vsip_cidr_t *cidr,
-						  __u64 bitmap);
+						  __u64 *bitmap);
```
h-w-chen: what is the reason to change this param to type of pointer?

src/cli/test/test_cli.c
```diff
@@ -2148,24 +2155,33 @@ static void test_trn_cli_update_transit_network_policy_subcmd(void **state)
...
+		.interface = itf,
+		.tunid = 1,
+		.local_ip = 0x100000a,
+		.cidr_prefixlen = 17,
+		.cidr_ip = 0x60000ac,
+		.cidr_type = 2,
+		.bit_val = 1
+	};
+
+	struct rpc_trn_vsip_cidr_t policies[2];
```
h-w-chen: nit: could have initialized elements on definition of array

h-w-chen (changes requested):
sizeof(pointer type) is not the same as sizeof(array)

src/dmn/trn_transit_xdp_usr.c
clu2xlu: I tried using int pointer, it did return the correct value. sizeof(ipcidr[0]) == sizeof(*ipcidr) == sizeof(struct vsip_cidr_t)
https://stackoverflow.com/questions/1898657/result-of-sizeof-on-array-of-structs-in-c

src/dmn/trn_transit_xdp_usr.h
clu2xlu: I need pass in an array instead of individual values

src/cli/test/test_cli.c
clu2xlu: will fix

src/dmn/trn_transit_xdp_usr.c
h-w-chen: the problem is like this: 
struct foo{...};
struct foo *pFoo;
sizeof(pFoo) is always the length of pointer (8 in my linux machine).

src/dmn/trn_transit_xdp_usr.h
h-w-chen: in order to pass an array, usually the counter nneds to be one of the params too.

src/dmn/trn_transit_xdp_usr.c
clu2xlu: sizeof(pFoo) is the length of pointer. sizeof(*pFoo) is the length of the first element. sizeof(pFoo)/sizeof(*pFoo) is the number of elements. 

int main()
{
    int chars[20];
    printf("size of *chars: %ld \n", sizeof(*chars));
    printf("size of chars: %ld \n", sizeof(chars));
    printf("size of int: %ld \n", sizeof(int));
    return 0;
}

this outputs:
size of *chars: 4                                                                                                                                                                              
size of chars: 80                                                                                                                                                                              
size of int: 4  
   
https://www.onlinegdb.com/online_c_compiler

h-w-chen (approved):
/lgtm

PR #330, opened 2020-12-28
conn_track: set connection state for traffic denied by policy instead of simple removal 

### LINKED ISSUE #312: stray reply packet may lead to reply flow inserted as initiator direction
When policy update happens and the original connection is removed from conn_track map, the reply packets, that possibly happens afterwards, will leads to a mistaken connection inserted in the map, which in turn makes policy check on the other direction allow the traffic pass s it seems the **reply of the connection just inserted**. 

Need to provide fix to plug the hole.

h-w-chen:
This fixes #312.

The new design of connection track requires the conn has status
| bit value | Description |
| ----------- | ----------- |
| 0x01      | traffic is denied or open (allowed)  |
| 0x04      | need to re-evaluate applicable policy, or simply base on decision already made before  |

When traffic is denied by policy enforcement, its tracked connection will have state 0x01 (for tcp), or 0x05 (udp); if allowed, 0x00(tcp) or 0x04(udp), respectively.

When xdp prog is making decision for the reply flow, for packets that are reply of a tracked connection, it will re-evaluate with the _derived_ policy - if the state indicates so (which applies to udp flows).

This PR changes the xdp prog with update to the conn state, particularly with the re-eval flag.

### DIFF
--- a/src/xdp/trn_agent_xdp.c
+++ b/src/xdp/trn_agent_xdp.c
@@ -335,10 +335,11 @@ static __inline int trn_process_inner_ip(struct transit_packet *pkt)
 
 	if (pkt->inner_ipv4_tuple.protocol == IPPROTO_TCP || pkt->inner_ipv4_tuple.protocol == IPPROTO_UDP) {
 		__u8 *tracked_state = get_originated_conn_state(&conn_track_cache, pkt->agent_ep_tunid, &pkt->inner_ipv4_tuple);
+		// todo: only check for bi-directional connections
 		if (NULL != tracked_state){
 			if (0 != egress_reply_packet_check(pkt->agent_ep_tunid, &pkt->inner_ipv4_tuple, *tracked_state))
 			{
-				bpf_debug("[Agent:%ld.0x%x] ABORTED: packet to 0x%x egress policy denied, reply of a denied UDP conn\n",
+				bpf_debug("[Agent:%ld.0x%x] ABORTED: packet to 0x%x egress policy denied, reply of a denied conn\n",
 					pkt->agent_ep_tunid,
 					bpf_ntohl(pkt->agent_ep_ipv4),
 					bpf_ntohl(pkt->inner_ip->daddr));
--- a/src/xdp/trn_agent_xdp.c
+++ b/src/xdp/trn_agent_xdp.c
@@ -333,29 +333,18 @@ static __inline int trn_process_inner_ip(struct transit_packet *pkt)
 		pkt->inner_ipv4_tuple.dport = pkt->inner_udp->dest;
 	}
 
-	// todo: add conn_track related logic properly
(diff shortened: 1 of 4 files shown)


src/xdp/trn_agent_xdp.c
```diff
@@ -333,29 +333,19 @@ static __inline int trn_process_inner_ip(struct transit_packet *pkt)
...
-					bpf_debug("[Agent:%ld.0x%x] ABORTED: packet to 0x%x egress policy denied, reply of a denied UDP conn\n",
-						pkt->agent_ep_tunid,
-						bpf_ntohl(pkt->agent_ep_ipv4),
-						bpf_ntohl(pkt->inner_ip->daddr));
-					conntrack_remove_tcpudp_conn(&conn_track_cache, pkt->agent_ep_tunid, &pkt->inner_ipv4_tuple);
-					return XDP_ABORTED;
-				}
+		__u8 *tracked_state = get_originated_conn_state(&conn_track_cache, pkt->agent_ep_tunid, &pkt->inner_ipv4_tuple);
+		// todo: only check for bi-directional connections
+		if (NULL != tracked_state){
```
clu2xlu, line 339: bpf map lookup returns a int value. If not found, it returns a negative value

src/xdp/trn_transit_xdp.c
```diff
@@ -446,30 +446,19 @@ static __inline int trn_process_inner_ip(struct transit_packet *pkt)
...
-						tunnel_id,
-						bpf_ntohl(pkt->inner_ipv4_tuple.daddr),
-						bpf_ntohl(pkt->inner_ipv4_tuple.saddr));
-					// todo: to change conn state if it is available instead of removing it from the map
-					conntrack_remove_tcpudp_conn(&conn_track_cache, tunnel_id, &originated_tuple);
-					return XDP_ABORTED;
-				}
+		__u8 *tracked_state = get_originated_conn_state(&conn_track_cache, tunnel_id, &pkt->inner_ipv4_tuple);
+		// todo: only check for bi-directional connections
+		if (NULL != tracked_state) {
```
clu2xlu, line 452: same

clu2xlu (changes requested):

src/xdp/trn_agent_xdp.c
h-w-chen: The bpf helper used here ``` void *(*bpf_map_lookup_elem)(void *map, void *key) ```;
the map lookup function returned a void* which points to the (arbitrary) map element value. If not found, it returns NULL. In other words, NULL != x indicates an element found in the map. 
This helper is different from the one that userspace prog uses (which seems to take 3 params, one of them is pointer to value).

src/xdp/trn_transit_xdp.c
h-w-chen: explained in the other thread.

clu2xlu (approved):

PR #332, opened 2020-12-29
Change to batch update for transit enforcement map updates

### LINKED ISSUE #294: Change the policy update to batch style to improve performance

clu2xlu:
This partially resolves #294

For network policy enforcement maps updates, we were doing one by one bpf map delete previously. To increase efficiency, we now switch to batch update


h-w-chen (commented):
lgtm

h-w-chen (approved):

PR #333, opened 2020-12-30
conntrack xdp prog refactored to remove goto etc

h-w-chen:
This PR is refactoring only, not for new new features.

It gets rid of goto statement; removes unneeded function body (conntrack_remove_conn); removes tweaks (like out-of-if-block) that were introduced to satifiy bpf verifier when -O3 ws the compiler optimization level (now it is -O1; most of the teaks are obsolete).


clu2xlu (approved):
LGTM

PR #334, opened 2020-12-30
Add agent network policy cidr delete

clu2xlu:
This PR partially resolves issue #270

It gets cli command input and trigger RPC call, then delete from BPF map for network policies on the agent side.


h-w-chen (approved):
lgtm

PR #335, opened 2020-12-30
design doc: tracked connection has state

h-w-chen:
This PR is design document only.

Connection trcked and used by policy enforcement prog has state. When policy update is to block the trffic, the tracked entry won't be simply removed, but its state updated to denied instead. This design change is to avoid [stray reply packet leading to adverse entry](#312). The implementation is through PR #330.

The design change is to introduce connection state, and update the connection state reflecting the policy change, so as to prevent stray reply packet from creating incorrect entry in the tracking map.


clu2xlu (approved):
LGTM

PR #338, opened 2020-12-30
Send networkpolicy data from operator to daemon

### LINKED ISSUE #337: Send networkpolicy data from operator to daemon

Hong-Chang:


mizar/obj/endpoint.py
```diff
@@ -364,3 +375,109 @@ def update_agent_substrate(self, ep, bouncer):
 
     def delete_agent_substrate(self, ep, bouncer):
         self.rpc.delete_agent_substrate_ep(ep, bouncer.ip)
+
+    def update_networkpolicy_per_endpoint(self, data):
+        if len(data["old"]) > 0:
+            self.delete_network_policy_ingress("no_except", data["old"]["ingress"]["cidr_table_no_except"])
+            self.delete_network_policy_ingress("with_except", data["old"]["ingress"]["cidr_table_with_except"])
```
Hong-Chang, line 382: Cathy, let's discuss how we will combine these into batch. And shall we combine delete and update into batch.

mizar/obj/endpoint.py
```diff
@@ -364,3 +375,109 @@ def update_agent_substrate(self, ep, bouncer):
 
     def delete_agent_substrate(self, ep, bouncer):
         self.rpc.delete_agent_substrate_ep(ep, bouncer.ip)
+
+    def update_networkpolicy_per_endpoint(self, data):
```
clu2xlu, line 379: Any reason you dont want include this in mizar/obj/networkpolicy.py?

clu2xlu (changes requested):

mizar/obj/endpoint.py
Hong-Chang: To calculate bitmap value for network policies, there is quantity limitation for policy. So we decided to limit data scope to be within an endpoint. There could be 1000 policies in a cluster, but we expect less than 100 policies for a certain endpoint. 
This function is to handle data for an endpoint. That's the reason putting the func here.

clu2xlu:
Approving for now to unblock the engineer. But we need revisit this and tracking this with issue #344

clu2xlu (approved):
Approving for now to unblock the engineer. But we need revisit this and tracking this with issue #344

PR #343, opened 2020-12-31
batch update for protocol-port network policy on the transit side

### LINKED ISSUE #294: Change the policy update to batch style to improve performance

clu2xlu:
This partially resolves #294

For network policy protocol-port maps updates, we were doing one by one bpf map update previously. To increase efficiency, we now switch to batch update

### DIFF
--- a/src/dmn/trn_transit_xdp_usr.c
+++ b/src/dmn/trn_transit_xdp_usr.c
@@ -783,14 +783,18 @@ int trn_delete_transit_network_policy_enforcement_map(struct user_metadata_t *md
 
 int trn_update_transit_network_policy_protocol_port_map(struct user_metadata_t *md,
 						        struct vsip_ppo_t *policy,
-						        __u64 bitmap)
+						        __u64 *bitmap,
+							int counter)
 {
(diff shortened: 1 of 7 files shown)


src/dmn/trn_transit_xdp_usr.c
```diff
@@ -783,14 +783,18 @@ int trn_delete_transit_network_policy_enforcement_map(struct user_metadata_t *md
...
+	for (int i = 0; i < counter; i++)
+	{
+		int err = bpf_map_update_elem(md->ing_vsip_ppo_map_fd, &policy[i], &bitmap[i], 0);
 
-	if (err) {
-		TRN_LOG_ERROR("Update Protocol-Port ingress map failed (err:%d).",
-				err);
-		return 1;
+		if (err) {
+			TRN_LOG_ERROR("Update Protocol-Port ingress map failed (err:%d).",
```
h-w-chen: nit: if more specific error (like the key, value) could be logged, it may help to identify the problem easilier.

h-w-chen (approved):
generally lgtm.

PR #348, opened 2021-01-04
Protocol port update on the agent side

clu2xlu:
This PR partially resolves issue #270

It gets cli command input and trigger RPC call, then updates BPF map for network policy protocol port maps on the agent side.


h-w-chen (approved):
LGTM

PR #354, opened 2021-01-05
Networkpolicy adds label_networkpolicies map into operator_store

### LINKED ISSUE #353: Networkpolicy adds label_networkpolicies map into operator_store

Hong-Chang:


mizar/obj/networkpolicy.py
```diff
@@ -26,28 +26,16 @@
 
 
 class NetworkPolicy:
-    def __init__(self, name, obj_api, opr_store, spec=None):
```
clu2xlu, line 29: Please consider add network policy's metadata / spec in NetworkPolicy object in the future. Seems like you are only using NetworkPolicy object to store list of endpoints associated with the Network policy. If so, you can just use operator store to store this information. Seems pointless to have network policy object

mizar/networkpolicy/networkpolicy_util.py
```diff
@@ -69,14 +72,19 @@ def handle_networkpolicy_create_update(self, name, pod_label_dict, policy_types)
 
                 ep.set_data_for_networkpolicy(data_for_networkpolicy)
                 ep.update_networkpolicy_per_endpoint(data_for_networkpolicy)
+                for label in data_for_networkpolicy["ingress"]["label_networkpolicies_map"]:
+                    networkpolicy_opr.store.add_label_networkpolicy_ingress(label, data_for_networkpolicy["ingress"]["label_networkpolicies_map"][label])
+                for label in data_for_networkpolicy["egress"]["label_networkpolicies_map"]:
```
clu2xlu, line 77: This can be converted into network policy object specs

clu2xlu (approved):
Due to the tight deadline, approving this PR for now. Engineer should consider adding data_for_networkpolicy info into network policy object (mizar/obj/networkpolicy.py)

mizar/obj/networkpolicy.py
Hong-Chang: Good idea. After planned "issues" done, will revisit this.

mizar/networkpolicy/networkpolicy_util.py
Hong-Chang: data_for_networkpolicy is not per networkpolicy. It's per endpoint. It is to calculate each endpoint's networkpolicy related data.

PR #336, opened 2020-12-30
Add agent enforcement map update

clu2xlu:
This PR partially resolves issue #270

It gets cli command input and trigger RPC call, then updates BPF map for network policy enforcement maps on the agent side.

PR #357, opened 2021-01-06
Add size limit to cli command update_network_policy_ingress

clu2xlu:
When doing batch update, the cli command has a limit of 2047. 
This PR is to break down the batch size so that it does not exceeds 2047 each update

### DIFF
--- a/mizar/common/common.py
+++ b/mizar/common/common.py
@@ -44,6 +44,12 @@ def run_cmd(cmd):
     returncode = result.returncode
     return (returncode, text)
 
+def count_characters(self, a_dict):
+    count = 0
+    for key, value in a_dict.items():
+        count += len(key)
+        count += len(str(value))
+    return count
 
 def get_iface_index(name, iproute):
     intfs = iproute.link_lookup(ifname=name)
--- a/mizar/common/common.py
+++ b/mizar/common/common.py
@@ -44,12 +44,6 @@ def run_cmd(cmd):
     returncode = result.returncode
     return (returncode, text)
 
-def count_characters(self, a_dict):
-    count = 0
-    for key, value in a_dict.items():
-        count += len(key)
-        count += len(str(value))
-    return count
 
 def get_iface_index(name, iproute):
     intfs = iproute.link_lookup(ifname=name)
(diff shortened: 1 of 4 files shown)


mizar/common/common.py
```diff
@@ -44,6 +44,12 @@ def run_cmd(cmd):
     returncode = result.returncode
     return (returncode, text)
 
+def count_characters(self, a_dict):
```
h-w-chen: appriciate the effort to limit the size of cli args. My concern is this calc is approximation; python has ulitities to get the precise of json string length - like json stringfy. You may want to use that instead.

h-w-chen (changes requested):
better to use precise method

mizar/common/common.py
clu2xlu: I will be using json.dumps instead of stringify : https://stackoverflow.com/questions/46227854/json-stringify-javascript-and-json-dumps-python-not-equivalent-on-a-list
because space also count as one character.

mizar/common/rpc.py
```diff
@@ -329,14 +329,13 @@ def update_network_policy_ingress(self, cidr_networkpolicy_list):
                 "cidr_type": cidr_networkpolicy.get_cidr_type_int(),
                 "bit_value": str(cidr_networkpolicy.policy_bit_value),
             }
-            item_len = count_characters(conf)
-            counter += item_len
+            item_len = len(json.dumps(conf)) + 1
+            counter = len(json.dumps(conf_list)) + item_len
             if (counter < CONSTANTS.MAX_CLI_CHAR_LENGTH):
                 conf_list.append(conf)
             if (counter + item_len > CONSTANTS.MAX_CLI_CHAR_LENGTH):
```
h-w-chen, line 336: counter seems here already with the addition of item_len; what is the consideration to add it ne more time?

clu2xlu: for line #336, it's basically checking if the command line size is going out of bound at the next loop. Since the size of the str(conf) is about the same size each loop, I am doing an approx. on the size of the command line if we are adding one more conf here. 
If we check only the current size and it's out of bound, it would be too late

h-w-chen: ok sounds reasonable

h-w-chen (approved):
LGTM

PR #363, opened 2021-01-07
Check size limit for network policy related map updates and deletes

clu2xlu:
When doing batch updates and deletes, the cli command has a limit of 2047.
This PR is to break down the batch size so that it does not exceeds 2047 each update and delete

### DIFF
--- a/mizar/common/common.py
+++ b/mizar/common/common.py
@@ -25,6 +25,7 @@ import logging
 import luigi
 import kopf
 import datetime
+import json
 import dateutil.parser
 from kubernetes import watch, client
 from ctypes.util import find_library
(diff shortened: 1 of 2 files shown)


mizar/common/rpc.py
```diff
@@ -380,7 +390,17 @@ def delete_network_policy_ingress(self, cidr_networkpolicy_list):
                 "cidr_type": cidr_networkpolicy.get_cidr_type_int(),
                 "bit_value": str(cidr_networkpolicy.policy_bit_value),
             }
-            conf_list.append(conf)
+            # +1 is for comma that will get added during json conversion
```
h-w-chen: this block of code aplitting list to batches not exceeding a certian number, seems to repeat quite some times. Is it possible to reduce code repetition and insease code reuse?

h-w-chen (changes requested):
please see whether it is possible in line with DRY.

mizar/common/common.py
```diff
@@ -382,3 +383,11 @@ def reset_param(param):
     param.extra = None
     param.return_message = None
     return param
+
+def network_policy_rpc_helper(conf, conf_list):
```
h-w-chen: suggest to name the function more specifically; its purpose seems to check whether it would exceed the max length __IF__ one more element appends. How about name it like "conf_list_has_max_elements"?

h-w-chen (changes requested):
please rename function to reflect its specific intention.

h-w-chen (approved):
lgtm

PR #364, opened 2021-01-08
Handling events: Delete NetworkPolicy (kopf)

### LINKED ISSUE #275: Handling events: Delete NetworkPolicy (kopf)

Hong-Chang:
Combine networkpolicy update and delete to be same code routine. The function has been integrated test well.

### DIFF
--- a/mizar/networkpolicy/networkpolicy_util.py
+++ b/mizar/networkpolicy/networkpolicy_util.py
@@ -30,26 +30,76 @@ logger = logging.getLogger()
 
 
 class NetworkPolicyUtil:
-    def handle_networkpolicy_create_update(self, name, pod_label_dict, policy_types):
+    def update_networkpolicy_endpoints_mapping_and_return_affected_endpoint_names(self, policy_name, pod_label_dict):
+        affected_endpoint_names = set()
+
+        endpoint_name_list = self.retrieve_endpoints_for_networkpolicy(policy_name, pod_label_dict)
+
+        if policy_name not in networkpolicy_opr.store.networkpolicy_endpoints_store:
+            networkpolicy_opr.store.networkpolicy_endpoints_store[policy_name] = set()
+
+        for endpoint_name in networkpolicy_opr.store.networkpolicy_endpoints_store[policy_name]:
+            if endpoint_name in endpoint_name_list:
+                endpoint_name_list.remove(endpoint_name)
+            else:                
+                affected_endpoint_names.add(endpoint_name)
+                ep = networkpolicy_opr.store.get_ep(endpoint_name)
+                if ep is not None:
+                    ep.remove_ingress_networkpolicy(policy_name)
+                    ep.remove_egress_networkpolicy(policy_name)
+
+        for endpoint_name in affected_endpoint_names:
+            networkpolicy_opr.store.networkpolicy_endpoints_store[policy_name].remove(endpoint_name)
+
+        for endpoint_name in endpoint_name_list:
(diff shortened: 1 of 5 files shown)


mizar/networkpolicy/networkpolicy_util.py
```diff
@@ -30,26 +30,76 @@
 
 
 class NetworkPolicyUtil:
-    def handle_networkpolicy_create_update(self, name, pod_label_dict, policy_types):
+    def update_networkpolicy_endpoints_mapping_and_return_affected_endpoint_names(self, policy_name, pod_label_dict):
```
h-w-chen: this func name seems too long. From the name, it seems to achieve two things - update the mapping, and get affected endpoints. If so, please consider split to 2 functions; otherwise, please consider renaming it to reflect its purpose.

mizar/networkpolicy/networkpolicy_util.py
```diff
@@ -30,26 +30,76 @@
...
         pods = kube_list_pods_by_labels(networkpolicy_opr.core_api, pod_label_dict)
         if pods is None:
-            return
+            return endpoint_name_list
+        
+        for pod in pods.items:
+            pod_name = pod.metadata.name
+            eps = endpoint_opr.store.get_eps_in_pod(pod_name)
+            for ep in eps.values():
+                endpoint_name_list.add(ep.name)
```
h-w-chen, line 79: not sure if pods in different namespace could be handled properly. Is the ep name has namespace as part of its name?

mizar/networkpolicy/networkpolicy_util.py
```diff
@@ -30,26 +30,76 @@
...
+            if ep is not None:
+                if has_ingress and policy_name not in ep.ingress_networkpolicies:
+                    ep.add_ingress_networkpolicy(policy_name)
+                if has_egress and policy_name not in ep.egress_networkpolicies:
+                    ep.add_egress_networkpolicy(policy_name)
+
+    def handle_networkpolicy_update_delete(self, endpoint_name_list):
+        for endpoint_name in endpoint_name_list:
+            ep = networkpolicy_opr.store.get_ep(endpoint_name)
+            if ep is not None:
```
h-w-chen: suggest to add logging when ep is none; my understanding is when it is none, something had happened between the ep name was got and when this func is being called, which is likey some race condition / exception.

h-w-chen (changes requested):
please consider either rename the long-named func or refactor to spilit, following single responsibility principle.

mizar/networkpolicy/networkpolicy_util.py
Hong-Chang: Fixed

mizar/networkpolicy/networkpolicy_util.py
Hong-Chang: Fixed

mizar/networkpolicy/networkpolicy_util.py
Hong-Chang: This is known issue: mizar doesn't handle namespace. Which means when the objects are in different namespaces, mizar just ignores namespace but only use name property. When there is same name objects in different namespaces, the system result is unexpected.
I created issue https://github.com/CentaurusInfra/mizar/issues/365 for tracking.

h-w-chen: OK got it. Thanks for opening the issue tracking it.

h-w-chen (approved):
lgtm

PR #366, opened 2021-01-09
basic network metrics of kindnet and cilium

### LINKED ISSUE #349: perf test - compare mizar network performance with other few significant k8s cni plugins (cilium etc)
target env: K8S installed via Kind
scope: 
* basic network metrics (throughput, latency)
* data plane only

h-w-chen:
This closes #349.

This PR is document only.

We had collected Mizar's basic network metrics with Kind 1+2 env. In order to evaluate in quantitative manner, we need to collect same set of metrics of other competing network providers, with the similar env setup. This PR recordes the metrics of Kindnet and Cilium reusing the same AWS Kind 1+2 env where we had tested Mizar.

### DIFF
--- a/docs/dev/network-perf-test.md
+++ b/docs/dev/network-perf-test.md
@@ -5,7 +5,7 @@ Also, it has one sample collection of network metrics for the reference, which w
 
 
 ## Background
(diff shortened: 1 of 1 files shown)


clu2xlu (approved):
LGTM

PR #368, opened 2021-01-09
Networkpolicy: when pod label update, it may affect endpoint related with policy. Handle the change

### LINKED ISSUE #281: Add NetworkPolicy to store

### LINKED ISSUE #362: Networkpolicy: when pod label update, it may affect endpoint related with policy. Handle the change

Hong-Chang:
Also added NetworkPolicy to operator store.

### DIFF
--- a/mizar/dp/mizar/workflows/builtins/networkpolicies/delete.py
+++ b/mizar/dp/mizar/workflows/builtins/networkpolicies/delete.py
@@ -37,9 +37,9 @@ class k8sNetworkPolicyDelete(WorkflowTask):
         logger.info("Run {task}".format(task=self.__class__.__name__))
 
         policy_name = "{}:{}".format(self.param.namespace, self.param.name)
-        affected_endpoint_names = networkpolicy_util.update_and_retrieve_affected_endpoint_names(policy_name, None, None)
+        endpoint_names = networkpolicy_util.update_and_retrieve_endpoint_names(policy_name, None, None)
 
-        networkpolicy_util.handle_networkpolicy_update_delete(affected_endpoint_names)
+        networkpolicy_util.handle_networkpolicy_update_delete(endpoint_names)
         networkpolicy_opr.store.delete_networkpolicy(policy_name)
 
         self.finalize()
--- a/mizar/networkpolicy/networkpolicy_util.py
+++ b/mizar/networkpolicy/networkpolicy_util.py
@@ -30,9 +30,11 @@ logger = logging.getLogger()
 
 
 class NetworkPolicyUtil:
-    def update_and_retrieve_affected_endpoint_names(self, policy_name, pod_label_dict):
+    def update_and_retrieve_affected_endpoint_names(self, policy_name, pod_label_dict, policy_types):
         affected_endpoint_names = set()
 
+        self.update_pod_label_networkpolicy_mapping_in_store(policy_name, pod_label_dict)
+        
         endpoint_name_list = self.retrieve_endpoints_for_networkpolicy(policy_name, pod_label_dict)
 
         if policy_name not in networkpolicy_opr.store.networkpolicy_endpoints_store:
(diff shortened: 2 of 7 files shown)


mizar/dp/mizar/workflows/builtins/networkpolicies/delete.py
```diff
@@ -35,9 +37,9 @@ def run(self):
         logger.info("Run {task}".format(task=self.__class__.__name__))
 
         policy_name = "{}:{}".format(self.param.namespace, self.param.name)
-        affected_endpoint_names = networkpolicy_util.update_and_retrieve_affected_endpoint_names(policy_name, None)
+        affected_endpoint_names = networkpolicy_util.update_and_retrieve_affected_endpoint_names(policy_name, None, None)
```
h-w-chen: is the purpose of this call to get ALL endpoints of the named policy? If so, suggest to make is clear to reader by, e.g. intruducing a small function named as "get_policy_endpoints".

mizar/networkpolicy/networkpolicy_util.py
```diff
@@ -345,10 +360,23 @@ def handle_pod_change_for_networkpolicy(self, diff):
...
             self.add_affected_networkpolicy_by_pod_label(policy_name_list, label)
-
         for label in data["remove"]:
             self.add_affected_networkpolicy_by_pod_label(policy_name_list, label)
 
+        endpoint_affected_policy_name_list = set()
+        for label in data["add"]:            
+            self.add_endpoint_affected_networkpolicy_by_pod_label(endpoint_affected_policy_name_list, label)
+        for label in data["remove"]:
+            self.add_endpoint_affected_networkpolicy_by_pod_label(endpoint_affected_policy_name_list, label)
```
h-w-chen, line 376: not sure "add_endpoint_..." is really intended or "remove_endpoint_...".  If so, please add comment to explain it a bit.

mizar/networkpolicy/networkpolicy_util.py
```diff
@@ -359,6 +387,11 @@ def add_affected_networkpolicy_by_pod_label(self, policy_name_list, label):
             for policy_name in networkpolicy_opr.store.label_networkpolicies_egress_store[label]:
                 policy_name_list.add(policy_name)
 
+    def add_endpoint_affected_networkpolicy_by_pod_label(self, policy_name_list, label):
```
h-w-chen, line 401: from the function body, it seems not very apparant the endpoint is involved semantically. Suggest to revise the function name.

mizar/obj/networkpolicy.py
```diff
@@ -0,0 +1,66 @@
+# SPDX-License-Identifier: MIT
```
h-w-chen, line 1: better in line with license of whole project

h-w-chen (changes requested):
1. please add a bit more content in PR description make the focus of PR clear to reviewers;
2. as a genral suggestion, please name function in line with the implentation and purpose.

mizar/dp/mizar/workflows/builtins/networkpolicies/delete.py
Hong-Chang: Changed affected_endpoint_names to endpoint_names.
There is networkpolicy_endpoints_store in operator store which stores mapping of policy to endpoints.
Here the function named update_and_retrieve_endpoint_names. Its name means by input of policy's pod selector, it will update networkpolicy_endpoints_store and return updated endpoint name list for later usage.

mizar/networkpolicy/networkpolicy_util.py
Hong-Chang: I added comments in code:
# Followed code piece is to calculate how many policy are affected by the label change, and form data of endpoint_affected_policy_name_list.
# Either label is added or removed, policy is affected, so for both data["add"] and data["remove"] we call same function add_endpoint_affected_networkpolicy_by_pod_label.
 # Here label is defined in policy's podSelector which affecting policy's mapping to endpoints.

mizar/networkpolicy/networkpolicy_util.py
Hong-Chang: Here endpoint is not a noun. It's adjectives to describe networkpolicy. It's "endpoint_affected_networkpolicy". 
In a networkpolicy definition, the podSelector is a list of labels. And the podSelector can appear two places: one is in policy ingress/egress rules, the second one is in the beginning of the policy definition which means the policy's power on the pods/endpoints. Accordingly, when a pod changed label, it may affect networkpolicy ingress/egress rules, or affect networkpolicy-endpoints mapping. If a policy is affected by pod label change, I call it "endpoint affected networkpolicy".

I added above comments into the code.

mizar/obj/networkpolicy.py
Hong-Chang: I see all the python code files are using SPDX-License-Identifier: MIT. 
Looks like c code are using SPDX-License-Identifier: GPL-2.0-or-later.
I created issue https://github.com/CentaurusInfra/mizar/issues/369 for tracking.

mizar/networkpolicy/networkpolicy_util.py
h-w-chen: Having comments explaining the intention is helpful, if the function needs clarification. good move.

h-w-chen (approved):
it seems fine to go.

PR #371, opened 2021-01-12
When deleting a networkpolicy, make sure all its references in all related maps are deleted

### LINKED ISSUE #367: When delete a networkpolicy, make sure all its references in all related maps are deleted

Hong-Chang:
Implementation:
1. When deleting a network policy, delete label_policy_store as well.
2. When deleting a network policy, remove the policy from label-policy mapping.

### DIFF
--- a/mizar/store/operator_store.py
+++ b/mizar/store/operator_store.py
@@ -205,6 +205,22 @@ class OprStore(object):
     def delete_networkpolicy(self, name):
         if name in self.networkpolicies_store:
             del self.networkpolicies_store[name]
+        self.delete_networkpolicy_in_label_policy_store(self.label_networkpolicies_store, name)
+        self.delete_networkpolicy_in_label_policy_store(self.label_networkpolicies_ingress_store, name)
+        self.delete_networkpolicy_in_label_policy_store(self.label_networkpolicies_egress_store, name)
+        self.delete_networkpolicy_in_label_policy_store(self.namespace_label_networkpolicies_ingress_store, name)
+        self.delete_networkpolicy_in_label_policy_store(self.namespace_label_networkpolicies_egress_store, name)
+
+    def delete_networkpolicy_in_label_policy_store(self, label_policy_store, policy_name):
+        label_list = set()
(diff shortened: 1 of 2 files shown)


mizar/store/operator_store.py
```diff
@@ -205,6 +205,22 @@ def update_networkpolicy(self, networkpolicy):
...
             del self.networkpolicies_store[name]
+        self.delete_networkpolicy_in_label_policy_store(self.label_networkpolicies_store, name)
+        self.delete_networkpolicy_in_label_policy_store(self.label_networkpolicies_ingress_store, name)
+        self.delete_networkpolicy_in_label_policy_store(self.label_networkpolicies_egress_store, name)
+        self.delete_networkpolicy_in_label_policy_store(self.namespace_label_networkpolicies_ingress_store, name)
+        self.delete_networkpolicy_in_label_policy_store(self.namespace_label_networkpolicies_egress_store, name)
+
+    def delete_networkpolicy_in_label_policy_store(self, label_policy_store, policy_name):
+        label_list = set()
+        for label in label_policy_store:
```
XiaoningDing, line 216: Is line 216 looping  through all labels and do a search for each label?  If yes it seems not efficient.  Can we first get labels from the policy, and then only remove from these labels?

XiaoningDing (approved):

mizar/store/operator_store.py
Hong-Chang: Good thoughts. Here actually we are considering two approaches: First one is get labels from policy-label mapping, Second is get labels from policy by on the fly querying the policy. We are in the trend to use First approach.

For now I haven't stored policy-label mapping. I want to do it when necessary. This pr's scenario happens when deleting a network policy. In such scenario, I feel it's ok to handle the deletion slowly. It's because it's rare to delete a policy. And it's rare to create a same name policy after delete it. So it's ok to slowly clean the data for this deleted policy.

In next release we will store more data into cache, and policy-label mapping is the one planned to be in the cache. At that time this function will be optimized.

PR #380, opened 2021-01-14
NetworkPolicy should only affect pods in the same namespace

### LINKED ISSUE #377: NetworkPolicy should only affect pods in the same namespace

Hong-Chang:
Implementation:
1. Add namespace field into networkpolicy object.
2. When query pods, add namespace parameter.

### DIFF
--- a/mizar/common/common.py
+++ b/mizar/common/common.py
@@ -289,12 +289,14 @@ def kube_list_namespaces_by_labels(core_api, label_dict):
     except:
         return None
 
-def kube_list_pods_by_labels(core_api, label_dict):
+def kube_list_pods_by_labels(core_api, label_dict, namespace=None):
     try:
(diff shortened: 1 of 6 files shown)


mizar/common/common.py
```diff
@@ -289,12 +289,14 @@ def kube_list_namespaces_by_labels(core_api, label_dict):
     except:
         return None
 
-def kube_list_pods_by_labels(core_api, label_dict):
+def kube_list_pods_by_labels(core_api, label_dict, namespace=None):
```
h-w-chen, line 292: what is the semantic meaning for none namespace? all namespaces?

h-w-chen (approved):
lgtm

mizar/common/common.py
Hong-Chang: Yes, when not sending namespace value, it means searching from all namespaces.

PR #381, opened 2021-01-14
eth0 workaround #251

### LINKED ISSUE #251: [BUG] Provide eth0 workaround

click2cloud-akshay:


mizar/get_interface.py
```diff
@@ -0,0 +1,11 @@
+import os
+import logging
+logger = logging.getLogger()
+
+def get_itf():
+    default_itf = os.popen("route | grep '^default' | grep -o '[^ ]*$'").read().split('\n')[0]
+    if "NET_ITF" in os.environ:
+        return os.getenv("NET_ITF")
```
clu2xlu: Can you change the name from "NET_ITF" to "MIZAR_ITF"? Because each machine has many environment variables, we want distinguish mizar specific env var from the reset

clu2xlu (approved):
looks good overall. Minor change on the naming.

mizar/get_interface.py
```diff
@@ -0,0 +1,11 @@
+import os
+import logging
+logger = logging.getLogger()
+
+def get_itf():
+    default_itf = os.popen("route | grep '^default' | grep -o '[^ ]*$'").read().split('\n')[0]
+    if "NET_ITF" in os.environ:
```
XiaoningDing: Suggest changing the environment variable with prefix "Mizar" to: 1) make it clear that it's for Mizar. 2) reduce the change of name conflicts.  Such as "MIZAR_NET_INTERFACE".

mizar/get_interface.py
```diff
@@ -0,0 +1,11 @@
+import os
+import logging
+logger = logging.getLogger()
+
+def get_itf():
+    default_itf = os.popen("route | grep '^default' | grep -o '[^ ]*$'").read().split('\n')[0]
```
XiaoningDing, line 6: Not sure if "route" command is always available. We probably can set default_itf to "eth0" if the execution and parsing result is empty?

XiaoningDing (approved):
Thanks, the PR looks good to me.  Added two comments for your consideration.

mizar/get_interface.py
click2cloud-akshay: We can set it to 'eth0' but, it will work again as current the scenario of the eth0 bug. As the interface name is generally changed with the version of Ubuntu OS. In ubuntu 16.0 it was 'eth0' but in ubuntu 18.0 and above the interface name changed to something like 'enXXX' due to the Predictable Interface Names.

click2cloud-akshay: We can set it to 'eth0' but, it will work again as current the scenario of the eth0 bug. As the interface name is generally changed with the version of Ubuntu OS. In ubuntu 16.0 it was 'eth0' but in ubuntu 18.0 and above the interface name changed to something like 'enXXX' due to the Predictable Interface Names.

mizar/get_interface.py
click2cloud-akshay: Updated the environment variable name as per the suggestion.

mizar/get_interface.py
click2cloud-akshay: Updated the environment variable name as per the suggestion.

PR #382, opened 2021-01-14
Change enforcement map update and delete from batches to singles

clu2xlu:
Enforcement map updates and deletes are different from other network policy related map updates and deletes. We decided to only use single update and delete for enforcement maps.


Hong-Chang (approved):
LGTM

PR #389, opened 2021-01-17
Fix eth error during enforcement map update and delete on the agent side

clu2xlu:
The agent enforcement map update failed is because wrong interface was picked (eth0 was picked instead of veth-xxxxxxx).


Hong-Chang (approved):
LGTM

Issue #396, opened 2021-01-19
NetworkPolicy Issue: delete_network_policy_protocol_port_ingress returns None Error


Hong-Chang:
Repro:
kubectl run pod0 --image=fwnetworking/testpod
kubectl run pod1 --image=fwnetworking/testpod

kubectl apply -f  /home/ubuntu/go/src/k8s.io/mizar/mizar/obj/tests/test_networkpolicy.yaml
kubectl apply -f /home/ubuntu/go/src/k8s.io/mizar/mizar/obj/tests/test_networkpolicy_default_allow_ingress.yaml
kubectl apply -f /home/ubuntu/go/src/k8s.io/mizar/mizar/obj/tests/test_networkpolicy_default_deny_ingress.yaml

kubectl logs mizar-operator-65d5dd779d-rprzp | grep "returns None Error" -A20 -B20

Got:
[2021-01-19 08:02:00,253] root                 [INFO    ] delete_network_policy_protocol_port_ingress: /trn_bin/transit -s 172.18.0.2  delete-net-policy-protocol-port-in -i eth0 -j '[{"tunnel_id": "1", "local_ip": "20.0.0.26", "protocol": "6", "port": "8000", "bit_value": "2"}, {"tunnel_id": "1", "local_ip": "20.0.0.26", "protocol": "6", "port": "5976", "bit_value": "2"}]'
[2021-01-19 08:02:00,269] root                 [INFO    ] delete_network_policy_protocol_port_ingress returns None Error: delete_transit_network_policy_protocol_port_1 fatal daemon error, see transitd logs for details.
main -s 172.18.0.2
Connecting to 172.18.0.2 using udp protocol.

[2021-01-19 08:02:00,270] root                 [INFO    ] delete_network_policy_protocol_port_egress: /trn_bin/transit -s 172.18.0.2  delete-net-policy-protocol-port-out -j '[{"tunnel_id": "1", "local_ip": "20.0.0.26", "protocol": "6", "port": "8000", "bit_value": "2"}]' -i 'veth-894cad92'
[2021-01-19 08:02:00,287] root                 [INFO    ] delete_network_policy_protocol_port_egress returns None Error: delete_agent_network_policy_protocol_port_1 fatal daemon error, see transitd logs for details.
main -s 172.18.0.2
Connecting to 172.18.0.2 using udp protocol.

clu2xlu:
The actual error log is in /var/log/syslog. Do you have the log? 
[...]

PR #400, opened 2021-01-20
Bug fix for protocol port map update and delete

clu2xlu:
This resolves issue #396 
The issue is with RPC:
int UPDATE_TRANSIT_NETWORK_POLICY_PROTOCOL_PORT(rpc_trn_vsip_ppo_t) = 27;
int DELETE_TRANSIT_NETWORK_POLICY_PROTOCOL_PORT(rpc_trn_vsip_ppo_key_t) = 28;
int UPDATE_AGENT_NETWORK_POLICY_PROTOCOL_PORT(rpc_trn_vsip_ppo_t) = 33;
int DELETE_AGENT_NETWORK_POLICY_PROTOCOL_PORT(rpc_trn_vsip_ppo_key_t) = 34;

The size of the param rpc_trn_vsip_ppo_key_t and rpc_trn_vsip_ppo_t is fixed. Thus, only the first item of the array was passed to daemon.

Because RPC has above restriction, I can only pass one param rpc_trn_vsip_ppo_key_t or rpc_trn_vsip_ppo_t item at once to the RPC call. Thus, this PR is to switch to single update at the RPC level

### DIFF
--- a/src/cli/test/test_cli.c
+++ b/src/cli/test/test_cli.c
@@ -2816,7 +2816,7 @@ static void test_trn_cli_update_transit_network_policy_protocol_port_subcmd(void
 		.tunid = 3,
 		.local_ip = 0x300000a,
 		.proto = 6,
-		.port = 6379,
+		.port = 60184,
 		.bit_val = 10
 	},
 	{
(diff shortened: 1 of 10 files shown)


Hong-Chang (approved):
LGTM

PR #412, opened 2021-01-25
Fix policy update issue

Hong-Chang:
Original implementation has issue. Now change it to be more clear: When updating policy, find added endpoint and removed endpoints and handle it.


clu2xlu (approved):
LGTM

PR #424, opened 2021-01-29
mizarcni bug fix: bind mount arbitrary netns path to whch ip netns expects

### LINKED ISSUE #404: mizar cni: fix bug of netns path handling
current cni code assumes netns path od /var/run/netns/xxx, which is not the case all the time

h-w-chen:
This fixes #404.

The current mizar cni uses ip netns to handle netns and assumes netns path starting with "/var/run/netns/". Many container runtimes don't provide netns arg in such form (e.g. containerd provides /proc/\<pid\>/ns/netns), which would break the mizar cni plugin. This PR fixes the issue by bind mounting netns path to file under /var/run/netns/

This fix is necessary due to mizarcni being python script, which relies on ip netns shell command heavily.

### DIFF
--- a/mizar/common/common.py
+++ b/mizar/common/common.py
@@ -19,6 +19,7 @@
 # TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR
 # THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
+import os
 import subprocess
 import ctypes
 import logging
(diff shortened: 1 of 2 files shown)


mizar/common/common.py
```diff
@@ -415,3 +415,10 @@ def conf_list_has_max_elements(conf, conf_list):
     if (counter + item_len > CONSTANTS.MAX_CLI_CHAR_LENGTH):
         return True
     return False
+
+def bindmount_netns(src_netns_path, dst_netns_path):
+    import os
```
XiaoningDing: Shall we move these "import" to the beginning to keep style consistent?

XiaoningDing (approved):

mizar/common/common.py
h-w-chen: will update as suggested.

PR #428, opened 2021-02-03
containerizes mizarcni installation

### LINKED ISSUE #383: package mizar cni in a self-contained image
current mizar cni is composed of various python scripts, and relies on various python libraries. In order to facilitate the mizar deployment, we need to simplify cni installation.

h-w-chen:
This closes #383.

Current mizar cni installation is through pre-built kindnode image in Kind dev env; this works for Kind as the node is using the customized image and has cni plugin embedded. For general cases like k8s by kubeadm, we need other means to have mizar cni plugin installed on nodes, preferably in containerized form. This PR is about the container image that accomplishes mizar cni plugin installation. 

Following command will install mizar cni plug-in onto the host (verified on Ubuntu 20.04.1 LTS):
```bash
docker run --privileged --network=host --pid=host -v /var:/install/var mizarnet/mizarcni:tagname 
```

### DIFF
--- a/etc/docker/mizarcni.Dockerfile
+++ b/etc/docker/mizarcni.Dockerfile
@@ -1,0 +1,23 @@
+# SPDX-License-Identifier: MIT
+# Copyright (c) 2020 The Mizar Authors.
+
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:The above copyright
+# notice and this permission notice shall be included in all copies or
+# substantial portions of the Software.THE SOFTWARE IS PROVIDED "AS IS",
+# WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED
+# TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+# NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE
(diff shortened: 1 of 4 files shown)


etc/docker/mizarcni.Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+# WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED
+# TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+# NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE
+# FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+# TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR
+# THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+FROM debian
+COPY . /var/mizar
+COPY etc/docker/install-mizarcni.sh /
```
clu2xlu: Is there any reason we need copy this file into root folder instead of /var/mizar ? 
running from root folder seems odd, maybe put it in ~/bin or /usr/local/bin or even /var/mizar.

etc/docker/mizarcni.Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+ENTRYPOINT ["bash"]
+CMD ["/install-mizarcni.sh"]
```
clu2xlu: What's the default permission of the file? My only concern is this file is executable. 
RUN chmod +x install-mizarcni.sh

clu2xlu (changes requested):

etc/docker/mizarcni.Dockerfile
h-w-chen: just wanted to stand that script out. In fact, we don't have to. I'll remove the copy command as suggested.

etc/docker/mizarcni.Dockerfile
h-w-chen: no need to explicitly chmod +x, as it runs by bash, the entrypoint command anyway.

clu2xlu (approved):
LGTM

PR #430, opened 2021-02-05
install mizar cni as part of node agent deployment

### LINKED ISSUE #405: mizar cni installation: make cni installed as part of mizar daemon deployment

h-w-chen:
This fixes #405.

Current mizar lacks a k8s way to install cni plugin. Kind env istalls mizar cni as pre-built work node image, which does not work for general k8s cluster setup. This PR puts mizarcni installation as part of mizar daemonSet.

This PR also has a minor fix of mizarcni code - to always initialize cni_args_dict. Since it is so trivial, it is included in this PR.


clu2xlu (approved):
LGTM

PR #432, opened 2021-02-09
Update setup-machine-arktos.sh

Hong-Chang:
setup-machine-arktos.sh is outdated. This pr is to update it by:
1. Remove unnecessary installs
2. Use go version go1.13.9
3. Stop using multi-tenant version of containerd

### DIFF
--- a/setup-machine-arktos.sh
+++ b/setup-machine-arktos.sh
@@ -87,13 +87,6 @@ GO111MODULE="on" go get sigs.k8s.io/kind@v0.7.0
 
 ####################
 
-echo Setup: Enlist Mizar
-
-cd ~
-git clone https://github.com/CentaurusInfra/mizar
-
-####################
-
 echo Setup: Mizar Related
 
 cd ~/mizar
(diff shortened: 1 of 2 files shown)


setup-machine-arktos.sh
```diff
@@ -185,18 +134,6 @@ pip3 install PyYAML
 
 ####################
 
-echo Setup: Change Containerd
```
h-w-chen, line 188: It seems that customized containerd with multi-tenant ability is still needed for arktos integration, for the current mizarcni implementation. You can have some explanation why the upstream containerd is used here as temporary placement.

h-w-chen (changes requested):
please add comments about customized containerd. It is fine not to use it right now; it should have some comment as reminder/explanation why such a temporary measure is taken.

setup-machine-arktos.sh
Hong-Chang: I filed issue https://github.com/CentaurusInfra/mizar/issues/433 for tracking.

h-w-chen (approved):
lgtm

PR #448, opened 2021-03-31
Network policy arktos mizar integration First Draft

clu2xlu:


docs/design/dp_network_policy_arktos_mizar_int.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Requirements
+* Support standard [network policy object spec] (https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#networkpolicy-v1-networking-k8s-io) in Arktos
+
+## Architecture
+
+Below is the overall architecture of the network policy support in Arktos and Mizar: 
+
+![alt text](png/network_policy_arktos_mizar_int_architecture.png "Architecture")
+
```
vinaykul, line 14: Isn't network policy controller part of the operator? From my understanding, operator = CRD + custom-controller. Please see: https://www.ibm.com/cloud/blog/kubernetes-operators , he gives a very good overview.
Does our current implementation work differently than this?

docs/design/dp_network_policy_arktos_mizar_int.md
```diff
@@ -0,0 +1,77 @@
...
+### Network Policy Create/Update Calculation
+
+Update_and_retrieve_endpoint_names: 
+
+1. Update operator store for label_networkpolicies_store[label] = network_policy which is stored in mizar operator store. Note that operator store is a in-memory storage system in mizar.
+2.	Based on label vs network_policy dictionary in step 1, find all the pod’s names that satisfy label key-value set. NOTE: this step involves calling Kubernetes API server to get list of pods.
```
vinaykul, line 26: Where in the code do we make this call today?

docs/design/dp_network_policy_arktos_mizar_int.md
clu2xlu: "Network Policy Controller" (in yellow) is referring mizar controller in arktos:  https://github.com/CentaurusInfra/arktos/tree/master/pkg/controller/mizar

"Network Policy Operator" (in red) is referring operator in mizar repo: 
https://github.com/CentaurusInfra/mizar/tree/dev-next/mizar/dp/mizar/operators

It's just a name, in mizar we call it operator, in arktos code repo we call it controller. 

Here is the grpc portion of the code in arktos: https://github.com/CentaurusInfra/arktos/blob/master/pkg/controller/mizar/grpc_adaptor.go

Here is the grpc server code in mizar: 
https://github.com/CentaurusInfra/mizar/blob/dev-next/mizar/arktos/arktos_service.py

Hope this helps. 

Mizar was implemented independently, mizar operator (is called controller in arktos) does the functions you mentioned CRD+ custom-controller. It's just that when we integrated with arktos, we need to have a watcher in arktos to trigger work flow/operator in mizar. The way we did that is through GRPC

docs/design/dp_network_policy_arktos_mizar_int.md
clu2xlu: https://github.com/CentaurusInfra/mizar/blob/dev-next/mizar/dp/mizar/workflows/builtins/networkpolicies/create.py#L43

docs/design/dp_network_policy_arktos_mizar_int.md
```diff
@@ -0,0 +1,77 @@
...
+4.	Add all endpoints retrieved in step 3 and add them to a list (ep_names_to_be_handled), and then remove endpoints that was in networkpolicy_endpoints_ingress_store or networkpolicy_endpoints_egress_store (this is stored in mizar operator store) but no longer in ep_names_to_be_handled list. 
+
+handle_networkpolicy_update_delete
+
+1.	It gets the list of endpoints that are affected by the network policy update or create (step 4 above), and then mizar operator tries to retrieve all the endpoint objects (stored in mizar operator store) from the list. 
+2.	In step 3 above, each affected endpoint obj keeps track of all the network policy associated with this endpoint obj. Thus, mizar operator can get a list of network policies that are associated with this endpoint object and calculate the bit map which will update ebpf map later. 
+
+## Design Consideration and Challenges  
+
+Previously, with kind cluster, mizar can call kubernete API server directly and access operator store at the same time. However, with arktos integration, arktos API server and network policy operator (in red) does not communication with each other directly.
```
vinaykul: kind is just a means to deploy a local cluster that simulates multiple nodes. If we make calls from Mizar to apiserver in kind, we should be able to make the same calls with arktos or upstream multinode k8s deployment. What calls do Mizar make to the APIserver? Where in the code do we make calls to apiserver?

docs/design/dp_network_policy_arktos_mizar_int.md
```diff
@@ -0,0 +1,77 @@
...
+handle_networkpolicy_update_delete
+
+1.	It gets the list of endpoints that are affected by the network policy update or create (step 4 above), and then mizar operator tries to retrieve all the endpoint objects (stored in mizar operator store) from the list. 
+2.	In step 3 above, each affected endpoint obj keeps track of all the network policy associated with this endpoint obj. Thus, mizar operator can get a list of network policies that are associated with this endpoint object and calculate the bit map which will update ebpf map later. 
+
+## Design Consideration and Challenges  
+
+Previously, with kind cluster, mizar can call kubernete API server directly and access operator store at the same time. However, with arktos integration, arktos API server and network policy operator (in red) does not communication with each other directly. 
+
+* The communication between arktos API server and mizar network policy operator is one directional. Mizar network policy operator cannot send requests back to arktos API server.
```
vinaykul: As I mentioned earlier, operator = contorller + CRD. A controller should be able to watch and perform get/post operations to APIserver. But I might be mistaken, let's go over this in the review today.

docs/design/dp_network_policy_arktos_mizar_int.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Design Consideration and Challenges  
+
+Previously, with kind cluster, mizar can call kubernete API server directly and access operator store at the same time. However, with arktos integration, arktos API server and network policy operator (in red) does not communication with each other directly. 
+
+* The communication between arktos API server and mizar network policy operator is one directional. Mizar network policy operator cannot send requests back to arktos API server. 
+* As mentioned in section “network policy create/update”, network policy creation process needs access both mizar operator store and arktos API server. If we were to move the entire network policy calculation process (currently managed by network policy operator in red) into network policy controller (in yellow), we can now send request to arktos API server. However, by doing so, we then lost access to mizar operator store and stored endpoint object information.
```
vinaykul, line 41: Mizar operator store is in-memory isn't it? Does that not move with the controller code?

docs/design/dp_network_policy_arktos_mizar_int.md
clu2xlu: https://github.com/CentaurusInfra/mizar/blob/dev-next/mizar/common/common.py#L282-L314

clu2xlu: Can we call aktos's API directly from mizar tho? I was told arktos does not support that

docs/design/dp_network_policy_arktos_mizar_int.md
clu2xlu: Let's talk about it during our meeting today. You have to consider mizar and arktos separately, the only media between them is GRPC.

docs/design/dp_network_policy_arktos_mizar_int.md
```diff
@@ -0,0 +1,77 @@
...
+### 1. Mizar Operator Store
+
+Add three additional object store into Mizar operator store to track pods and namespaces: 
+* node_pods_store: key = node's host ip address & value = names of the pods on each node
+* pod_label_store: key = pod name & value = each pod's labels
+* namespace_label_store: key = namespace name & value = each namespace's labels
+
+Are we going to have race condition here??? How to avoid race condition
+
+### 2. Add Network Policy Controller in Arktos Repo
```
vinaykul, line 63: What's the code that is currently in arktos repo under pkg/controller/mizar? Should that code even be there?

vinaykul (commented):
Thanks for working on this. Please see my comments. Thanks,

docs/design/dp_network_policy_arktos_mizar_int.md
clu2xlu: Under pkg/controller/mizar has all the controller code (in yellow). We need the code in order to watch events

docs/design/dp_network_policy_arktos_mizar_int.md
```diff
@@ -0,0 +1,77 @@
...
+
+* Watch network policy events such as create, update and delete
+* Compose network policy's metadata information into GRPC message and send through GRPC client 
+
+### 3. Add Namespace Controller in Arktos Repo (C2C)
+
+* Watch namespace events such as create, update and delete
+* Compose network policy's metadata information into GRPC message and send through GRPC client
```
phudtran, line 71: Can you give a an example/template of the gRPC message?

docs/design/dp_network_policy_arktos_mizar_int.md
```diff
@@ -0,0 +1,77 @@
...
+
+### 3. Add Label Information into GRPC messages for pods and namespaces (C2C)
+* List all labels' key/value sets into GRPC message for pod events
```
phudtran, line 80: Example of gRPC message

docs/design/dp_network_policy_arktos_mizar_int.md
```diff
@@ -0,0 +1,77 @@
...
+
+### 4. Remove all kubenetes API Calls During Network Policy Calculation 
+* Remove kubernetes API calls
```
phudtran, line 93: Will these calls need to be made on the Arktos side?

docs/design/dp_network_policy_arktos_mizar_int.md
phudtran: Mizar can reply to to Arktos gRPC client, it just can't star the conversation :)

phudtran (changes requested):

Hong-Chang (approved):
Had an offline discussion with Cathy. In general I have no problem on the design.

docs/design/dp_network_policy_arktos_mizar_int.md
phudtran: I don't think you don't need to remove anything on the Mizar side, we still need these API calls when running on vanilla Kubernetes right?

We simply need to restructure the network policy operator in way such that it can function with the new gRPC call.

If there's a function that makes an API call in the Mizar code during network policy creation, we can simply overload that function with an additional flag, or create a separate function to distinguish for Arktos.

phudtran (changes requested):

docs/design/dp_network_policy_arktos_mizar_int.md
phudtran: Discussed offline. The API call being made is simple and can be kept track of in the operator store

phudtran (approved):

PR #463, opened 2021-04-07
Implement Label Based Network Policies - Send label change XDP table

### LINKED ISSUE #450: Implement Label Based Network Policies - Send label change to daemon

### LINKED ISSUE #451: Implement Label Based Network Policies - Add label to Endpoint XDP table and Save it

Hong-Chang:
The changes are:
1. Add label values to endponts
2. Send label values to daemon via rpc
3. Add label values attributes to struct endpoint_t
4. Save label values to endpoint XDP table


mizar/common/rpc.py
```diff
@@ -134,7 +134,9 @@ def update_ep(self, ep):
             "mac": ep.get_mac(),
             "veth": ep.get_veth_name(),
             "remote_ips": ep.get_remote_ips(),
-            "hosted_iface": peer
+            "hosted_iface": peer,
+            "pod_label_value": ep.get_pod_label_value(),
```
phudtran, line 138: I feel like we're updating a lot of extraneous information here.
Why not create a separate RPC with pod_label_value, and namespace_label_value?
You can associate it with an endpoint by using the same key as update_ep (tunnel_id, and ep_ip).

So the json here would look like this instead
```
def update_network_label_policy:
    jsconf = {
        "tunnel_id": ep.get_tunnel_id(),
        "ip": ep.get_ip(),
        "hosted_iface": peer,
        "pod_label_value": ep.get_pod_label_value()
    }
```
You will also need to create a new BPF map
What are your thoughts?

phudtran (changes requested):

mizar/common/rpc.py
Hong-Chang: I think it will be easier to keep the label values are property of endpoint. Otherwise I have to copy all the routine including create object, sending to daemon, sending to new XDP table, query value from table and pass it as a separate parameter when the value is needed for outgoing traffic. In current implementation, once the label value is in XDP table, I will automatically have it for outgoing traffic because at that time there is agent meta data, and inside there is endpoint object, and inside there is label value.

phudtran: @clu2xlu and @vinaykul  What are your thoughts?

clu2xlu: I am trying to understand why keep the label value as part of update_ep is easier.... 
To me have only update_ep versus have update_ep AND update_network_label_policy do not change the complexity. 
Having update_ep AND update_network_label_policy should just be an addition call to rpc. Having label be a separate rpc call creates a separation of the hardware layer from the software layer of the OSI model.

vinaykul: @phudtran If I'm not mistaken, as far as data goes, these labels are fields carried in the Geneve header the packet we transmit. Are we saying struct endpoint_t is not the best place to store this information?

vinaykul: In today's sync meeting we discussed the idea of creating a separate generic struct to hold label values which I could leverage in the future for Bandwidth QoS project (need to mark pod as highPriority=true|false) based on pod annotation that the XDP program will use to classify a pod's traffic for rate-limiting if false)

PR #469, opened 2021-04-15
Use mizarnet docker account instead of fwnetworking. Allow non-root users to run kind-setup. Misc cleanups

vinaykul:
This fixes the issue USTC folks hit with kind-setup script by switching from fwnetworking account to mizarnet docker account. 

This PR also enables kind-setup to work for non-root users, and contains a few misc cleanups.

### DIFF
--- a/k8s/kind/create_cluster.sh
+++ b/k8s/kind/create_cluster.sh
@@ -29,18 +29,18 @@ NODES=${3:-1}
 reg_name='local-kind-registry'
 reg_port='5000'
 reg_network='kind'
-running="$(sudo docker inspect -f '{{.State.Running}}' "${reg_name}" 2>/dev/null || true)"
+running="$(docker inspect -f '{{.State.Running}}' "${reg_name}" 2>/dev/null || true)"
 reg_url=$reg_name
-kind_version=$(sudo kind version)
+kind_version=$(kind version)
 if [ "${running}" != 'true' ]; then
-  sudo docker run \
+  docker run \
     -d --restart=always -p "${reg_port}:5000" --name "${reg_name}" \
     registry:2
 fi
 
 case "${kind_version}" in
   "kind v0.7."* | "kind v0.6."* | "kind v0.5."*)
-    reg_url="$(sudo docker inspect -f '{{.NetworkSettings.IPAddress}}' "${reg_name}")"
+    reg_url="$(docker inspect -f '{{.NetworkSettings.IPAddress}}' "${reg_name}")"
     reg_network='bridge'
     ;;
 esac
@@ -67,7 +67,7 @@ do
 done
 
 # create a cluster with the local registry enabled in containerd and n Nodes.
-cat <<EOF | sudo kind create cluster --name kind --kubeconfig ${KINDCONF} --config=-
+cat <<EOF | kind create cluster --name kind --kubeconfig ${KINDCONF} --config=-
 kind: Cluster
 apiVersion: kind.x-k8s.io/v1alpha4
 networking:
@@ -80,8 +80,8 @@ ${FINAL_NODES}
 EOF
 
 if [[ $reg_network == "kind" ]]; then
-  sudo docker network connect $reg_network "${reg_name}"
-  for node in $(sudo kind get nodes); do
+  docker network connect $reg_network "${reg_name}"
+  for node in $(kind get nodes); do
     kubectl annotate node "${node}" "kind.x-k8s.io/registry=localhost:${reg_port}";
   done
 fi
--- a/k8s/kind/create_cluster.sh
+++ b/k8s/kind/create_cluster.sh
@@ -82,6 +82,6 @@ EOF
 if [[ $reg_network == "kind" ]]; then
   sudo docker network connect $reg_network "${reg_name}"
   for node in $(sudo kind get nodes); do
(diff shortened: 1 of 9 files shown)


kind-setup.sh
```diff
@@ -25,7 +25,7 @@
 function get_status() {
     OBJECT=$1
 
-    kubectl get $OBJECT 2> /tmp/kubetctl.err | awk '
+    sudo kubectl get $OBJECT 2> ${KUBECTL_LOG} | awk '
```
Hong-Chang: Currently in my environment "./kind-setup dev" works pretty well for latest code. I have concern after this pr, then all of our kubectl cmd needs to be under sudo?

vinaykul: ./kind-setup dev does not currently work if you are logged in as 'ubuntu' user. Are you logged in as root?

Hong-Chang: I'm using ubuntu. It's working fine.

phudtran: Yeah it works for me too.
I think we should leave the sudo as a preference for user rather than put it in the script.
Over on Zeta we actually disallow all our scripts from running as sudo altogether.

k8s/kind/create_cluster.sh
```diff
@@ -80,8 +80,8 @@ ${FINAL_NODES}
 EOF
 
 if [[ $reg_network == "kind" ]]; then
-  docker network connect $reg_network "${reg_name}"
-  for node in $(kind get nodes); do
+  sudo docker network connect $reg_network "${reg_name}"
+  for node in $(sudo kind get nodes); do
```
Hong-Chang: May check whether sudo is necessary for kind? For me it's not needed.

kind-setup.sh
vinaykul: Thanks for the reviews. I have revised the change to remove sudo from all kubectl commands - it was not required. I've left it for docker and kind commands. The reason is that non-root user may not be auto-added to the docker group, and  will get permission denied errors unless you add the user to docker group and re-login.

Hong-Chang: I'm not sure whether kind permission is related with docker. I have left another comments on the kind.

vinaykul: So all of you are able to follow the instructions https://github.com/CentaurusInfra/mizar/blob/dev-next/docs/user/getting_started.md and it works for you out-of-box? 

I just tried a fresh 20.04 ubuntu VM in AWS, did bootstrap, and did ./kind-setup dev and hit various issues. Is it only me?

vinaykul: This is what I get after following our getting_started guide:

ubuntu@ip-192-168-1-184:~/mizar$ ./kind-setup.sh dev
./kind-setup.sh: line 66: kind: command not found
Got permission denied while trying to connect to the Docker daemon socket at unix:///var/run/docker.sock: Post http://%2Fvar%2Frun%2Fdocker.sock/v1.24/networks/create: dial unix /var/run/docker.sock: connect: permission denied
Got permission denied while trying to connect to the Docker daemon socket at unix:///var/run/docker.sock: Post http://%2Fvar%2Frun%2Fdocker.sock/v1.24/build?buildargs=%7B%7D&cachefrom=%5B%5D&cgroupparent=&cpuperiod=0&cpuquota=0&cpusetcpus=&cpusetmems=&cpushares=0&dockerfile=k8s%2Fkind%2FDockerfile&labels=%7B%7D&memory=0&memswap=0&networkmode=default&rm=1&shmsize=0&t=localhost%3A5000%2Fkindnode%3Alatest&target=&ulimits=null&version=1: dial unix /var/run/docker.sock: connect: permission denied
k8s/kind/create_cluster.sh: line 34: kind: command not found
docker: Got permission denied while trying to connect to the Docker daemon socket at unix:///var/run/docker.sock: Post http://%2Fvar%2Frun%2Fdocker.sock/v1.24/containers/create?name=local-kind-registry: dial unix /var/run/docker.sock: connect: permission denied.
See 'docker run --help'.
...
...

phudtran: This is what I did to get it to work as a non-root user
https://docs.docker.com/engine/install/linux-postinstall/

Hong-Chang: I know to avoid sudo for docker, there needs to be groupadd for docker. And it's fair we don't assume user will do that. I'm just not sure whether we really need sudo for kind.

vinaykul: Ok, so bottomline is that our instructions does not work for non-root user out of box and results in poor experience. They need enough familiarity with linux to take these additional steps to get things to work. Is that a correct assessment?

vinaykul: And I also see that our documentation at readthedocs suggests using sudo, and we don't do so in our github docs. So we are inconsistent. Please see: https://mizar.readthedocs.io/en/latest/user/getting_started/

We should pick one route and be consistent about it. Let's discuss tomorrow.

k8s/kind/create_cluster.sh
vinaykul: I've updated the bootstrap script to get the user into docker group. This lets the user have permissions to access docker without sudo. Please review the latest update.

kind-setup.sh
vinaykul: I'm fixed up the bootstrap script which we ask folks to run to handle these hidden steps. Please review.

Hong-Chang (approved):
LGTM

phudtran (approved):
Looks good to me! Thanks Vinay!

PR #470, opened 2021-04-16
Add network policy to grpc for arktos integration

clu2xlu:


mizar/proto/mizar/proto/builtins.proto
```diff
@@ -68,6 +73,12 @@ message BuiltinsArktosMessage {
   string vpc = 2;
 }
 
+message BuiltinsNetworkPolicyMessage {
+  string name = 1;
+  string namespace = 2;
+  string tenant = 3;
```
Hong-Chang, line 79: Whether network policy has tenant field?

mizar/arktos/arktos_service.py
```diff
@@ -193,6 +211,15 @@ def DeleteService(self, request, context):
...
         return run_arktos_workflow(wffactory().k8sServiceDelete(param=param))
 
+    def DeleteNetworkPolicy(self, request, context):
+        logger.info(
+            "Deleting network policy from Network Controller {}".format(request.name))
+        param = reset_param(HandlerParam())
+        param.name = request.name
+        param.body['metadata'] = {}
+        param.body["metadata"]["namespace"] = request.namespace
+        param.extra = request.tenant
```
vinaykul, line 221: In create, we specify param.body['metadata']['tenant'] = request.tenant. Why is it different here?

mizar/proto/mizar/proto/builtins.proto
clu2xlu: Yes, it should

Hong-Chang (approved):
LGTM

PR #482, opened 2021-05-12
Mizar Stabilization and bug Fixes

phudtran:
This PR fixes bugs relating to the Mizar single yaml install file
#473
#474 
#475


vinaykul:
I noticed a few issues trying it out (besides the conflict). 
Seen this with kind-setup dev? 
```
688e187d6c79: Pushed 
latest: digest: sha256:9b6ac45c719d9b68fbf82a786d3cbb18dd36c226a5f74c324f46abf113d66d24 size: 4107
RTNETLINK answers: No such file or directory
Error: Nexthop has invalid gateway.
RTNETLINK answers: No such file or directory
Error: Nexthop has invalid gateway.
Waiting for cluster to come up..........................
```

vinaykul:
> I noticed a few issues trying it out (besides the conflict).
> Seen this with kind-setup dev?
> 
> ```
> 688e187d6c79: Pushed 
> latest: digest: sha256:9b6ac45c719d9b68fbf82a786d3cbb18dd36c226a5f74c324f46abf113d66d24 size: 4107
> RTNETLINK answers: No such file or directory
> Error: Nexthop has invalid gateway.
> RTNETLINK answers: No such file or directory
> Error: Nexthop has invalid gateway.
> Waiting for cluster to come up..........................
> ```

Never mind. This is my issue.

mizar/operator.py
```diff
@@ -113,11 +113,11 @@ def create_config_map():
         namespace="default",
         labels=dict(service="mizar")
     )
-    data = dict(host=socket.gethostbyname(socket.gethostname()))
+    host_ip = check_output(['hostname', '-I']).decode().split(" ")[0]
```
clu2xlu, line 116:
```
❯ hostname
ip-172-31-17-120
```
This is correct on aws machine, but it might not be the case in other ubuntu machines

clu2xlu: why we cannot use socket?

clu2xlu (changes requested):

mizar/operator.py
phudtran: It will work on any Linux machine, just not on MacOS and probably not Windows.
The reason I changed it was because in our lab environment, there is an issue with the socket.gethostbyname call when running it inside of the operator pod. 
I would prefer to use socket but I'm not sure why it doesn't work in this case, do you have any suggestions?

image

clu2xlu:
```
❯ kubectl exec -it mizar-operator-65d5dd779d-8wspj /bin/bash
kubectl exec [POD] [COMMAND] is DEPRECATED and will be removed in a future version. Use kubectl kubectl exec [POD] -- [COMMAND] instead.
root@kind-worker:/# python
Python 3.8.1 (default, Feb  2 2020, 08:49:34)
[GCC 8.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import socket
>>> socket.gethostbyname(socket.gethostname())
'172.18.0.5'
```

It's working for me acctualy. 

Usually gethostname() is getting information from /etc/hosts file (just my guess not 100% sure) If it's now working for you, you might want check if /etc/hosts file has correct entry

```
root@kind-worker:/# cat /etc/hosts
127.0.0.1       localhost
::1     localhost ip6-localhost ip6-loopback
fe00::0 ip6-localnet
ff00::0 ip6-mcastprefix
ff02::1 ip6-allnodes
ff02::2 ip6-allrouters
172.18.0.5      kind-worker
```

phudtran: Kind is ok, along with a AWS setup with Kubeadm 
Unfortunately our physical machines in the office are the ones having the issue

Here's the host file
It looks ok to me
```
127.0.0.1	localhost.localdomain	localhost
::1		localhost6.localdomain6	localhost6

# The following lines are desirable for IPv6 capable hosts
::1     localhost ip6-localhost ip6-loopback
fe00::0 ip6-localnet
ff02::1 ip6-allnodes
ff02::2 ip6-allrouters
ff02::3 ip6-allhosts```

Comment (clu2xlu) [reply mizar/operator.py:116]: we need edit this file before using socket.gethostname()

Comment (clu2xlu) [reply mizar/operator.py:116]: it's missing 172.18.0.5      kind-worker

Comment (clu2xlu) [reply mizar/operator.py:116]: The output of hostname is depend on what's in file /etc/hostname 
I tried on gcp ubuntu machine, the output is: 
```
ubuntu@pdc-stage-gke-bastion-us-east4:~$ hostname
pdc-stage-gke-bastion-us-east4
ubuntu@pdc-stage-gke-bastion-us-east4:~$ cat /etc/hostname
pdc-stage-gke-bastion-us-east4
```
Hostname is very customizable, and it's almost always will be customized by companies. 
In gcp, when you create an instance, you have to give it a name, and that's going to be what's in /etc/hostname hence the output of hostname. 

On a fresh ubuntu machine (neither aws nor gcp), the hostname is ubuntu

phudtran: Yeah, that string is already set.
Issue seems to lie elsewhere for these lab machines.
I think hostname command should be fine, since we will most likely not be running Mizar on a windows or macOS host.
I can send you the login for the machine if you'd like to dig a bit deeper too.

vinaykul: Which lab machine do you see problems with?

vinaykul:
The PR looks good to me with kind-setup.
I think we should merge this and then continue addressing any issues. I'm wondering if we should use getaddrinfo instead of gethostbyname. What do you guys think?

vinaykul:
LGTM

vinaykul (approved):

mizar/operator.py
clu2xlu: AWS machine will work fine. If you try on GCP machine, you will see a problem. All AWS EC2 machine has ip address listed as its hostname is because Amazon customized it. Same as GCP, google customized as user-input-hostname-region. 
We can't rely on customer to NOT change the hostname of their machine. Hope this make sense

phudtran: It's `hostname -I` not just `hostname`

clu2xlu: ok. that's fine then. Sorry for the confusion.

clu2xlu (approved):

vinaykul:
I tried this patch on a freshly created ubuntu 18.04 vm. Did bootstrap to update the kernel, built and setup a two-worker kind cluster, tried ping and UDP iperf tests. Looks good. 
We can address any desired changes regarding how host IP is queried in subsequent changes. I'll merge this PR.

PR #487, opened 2021-05-20
Modify arktos mizar deployment scripts

clu2xlu:


vinaykul:
Do we still need this PR in its current form now that we have merged Amit's PR?

clu2xlu:
> Do we still need this PR in its current form now that we have merged Amit's PR?

Yes. This change is based on Amit's PR

replace_containerd_start_arktos_api.sh
```diff
@@ -0,0 +1,13 @@
+cd $HOME
+wget -qO- https://github.com/futurewei-cloud/containerd/releases/download/tenant-cni-args/containerd.zip | zcat > /tmp/containerd
+sudo chmod +x /tmp/containerd
+sudo systemctl stop containerd
+sudo mv /usr/bin/containerd /usr/bin/containerd.bak
+sudo mv /tmp/containerd /usr/bin/
+sudo systemctl restart containerd
+sudo systemctl start docker
+export CONTAINER_RUNTIME_ENDPOINT="containerRuntime,container,/run/containerd/containerd.sock"
+sleep 15
```
vinaykul: Why do we need 15 second sleep here?

replace_containerd_start_arktos_api.sh
```diff
@@ -0,0 +1,13 @@
+cd $HOME
+wget -qO- https://github.com/futurewei-cloud/containerd/releases/download/tenant-cni-args/containerd.zip | zcat > /tmp/containerd
+sudo chmod +x /tmp/containerd
+sudo systemctl stop containerd
+sudo mv /usr/bin/containerd /usr/bin/containerd.bak
+sudo mv /tmp/containerd /usr/bin/
+sudo systemctl restart containerd
+sudo systemctl start docker
```
vinaykul, line 8: Why is this needed?

clu2xlu: We need replace arktos containerd with a modified version in order for mizar to work. If we dont do this, arktos does not pass tenant info to mizar

replace_containerd_start_arktos_api.sh
clu2xlu: Because it takes a while for containerd to restart

vinaykul: What happens if it took 16 seconds to restart?

replace_containerd_start_arktos_api.sh
vinaykul: What does that statement do if docker is already running? Also, are we now not using containerd instead of docker?

vinaykul:
+Yunwen to review this to see if we need to consider VM runtime as well in this workflow.

replace_containerd_start_arktos_api.sh
yb01: agree with Vinay, probably a better way to handle this is to check containerd status to ensure it is up or timeout with failure

yb01 (approved):
/lgtm, from vm runtime's perspective.

replace_containerd_start_arktos_api.sh
vinaykul: Thanks Yunwen. @clu2xlu We should avoid such arbitrary timeouts as it is going to fail at some point, leading to random/intermittent failures that are very hard to debug. Please fix this to check for desired status. For e.g. I did something similar for node ready in https://github.com/CentaurusInfra/mizar/blob/dev-next/kind-setup.sh#L116

vinaykul (changes requested):
Please get rid of arbitrary sleeps in the code that are destined to introduce intermittent failures that are hard to debug.

PR #490, opened 2021-05-25
Implement Label Based Network Policies - Judge traffic pass or block

### LINKED ISSUE #457: Implement Label Based Network Policies - Judge traffic pass or block

Hong-Chang:
What does this PR do?

1. Break original label defined network policies
2. Retrieve label values from packet. (Move the function to more proper place)
3. Check from label-policy xdp table whether the rule allows the traffic based on label values
4. Checking policy is in "shortcut" mode. Pod, namespace, pod+namespace policy will be check sequentially. Will return once matching the policy.

Why is it needed?
After the pr, the traffic will be judged pass or block by label based policy.

How was this tested?
I did manual e2e test and the traffics were judged as expected. 

No.

Closed #457


Hong-Chang:
Replaced by pr 493.

PR #492, opened 2021-05-27
Network bandwidth rate-limiting via EDT for low priority Pods

vinaykul:
What does this change do?
This PR adds feature that allows classification of Pods into high and low priority pods for network bandwidth, and uses the EDT algorithm to rate-limit traffic for low priority pods.

Why is it needed?
It partially implements the feature request to ensure high priority Pods have sufficient network bandwidth available to them.
See [design doc](https://docs.google.com/document/d/1zpmKGAYlm5V6M1yfNU0ES_Af1t8YM69oUJrow2sJtu0/edit).

How was this tested?
Manual E2E testing. (Unit tests and automated tests is work in progress)

No.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -98,7 +98,10 @@ clean::
 	rm -f *.gcov
 
 .PHONY: test
-test:: lcov functest
+test:: test_docker_image lcov functest
+
+test_docker_image:
+	sudo docker build -f ./test/Dockerfile -t buildbox:v2 ./test
 
 .PHONY: unittest
 unittest::
--- a/kind-setup.sh
+++ b/kind-setup.sh
@@ -126,7 +126,6 @@ source install/create_crds.sh $CWD
 source install/create_service_account.sh $CWD
 
 source install/deploy_daemon.sh $CWD $MODE $DOCKER_ACC
-sleep 5
 source install/deploy_operator.sh $CWD $MODE $DOCKER_ACC
 source install/create_testimage.sh $CWD $MODE $DOCKER_ACC
 
--- a/kind-setup.sh
+++ b/kind-setup.sh
@@ -63,7 +63,7 @@ KINDUSERCONF="${KINDUSERCONFDIR}/config"
 KUBECTL_LOG="/tmp/${USER}_kubetctl.err"
 MODE=${1:-user}
 NODES=${2:-3}
-timeout=240
+timeout=300
 
 if [[ "$MODE" == "dev" ]]; then
     DOCKER_ACC="localhost:5000"
@@ -113,10 +113,20 @@ api_ip=`docker inspect -f '{{range .NetworkSettings.Networks}}{{.IPAddress}}{{en
 sed "s/server: https:\/\/127.0.0.1:[[:digit:]]\+/server: https:\/\/$api_ip:6443/" $KINDCONF > $MIZARCONF
 ln -snf ${KINDCONF} ${KINDUSERCONF}
 
+while : ; do
+    kubectl get no -owide | grep NotReady > /dev/null
+    if [ $? -ne 0 ]; then
+        break
+    fi
+    echo "Waiting for node ready.."
+    sleep 10
+done
+
 source install/create_crds.sh $CWD
 source install/create_service_account.sh $CWD
 
 source install/deploy_daemon.sh $CWD $MODE $DOCKER_ACC
+sleep 5
 source install/deploy_operator.sh $CWD $MODE $DOCKER_ACC
 source install/create_testimage.sh $CWD $MODE $DOCKER_ACC
 
--- a/mizar/dp/mizar/workflows/builtins/pods/create.py
+++ b/mizar/dp/mizar/workflows/builtins/pods/create.py
@@ -95,6 +95,21 @@ class k8sPodCreate(WorkflowTask):
         ip = n.allocate_ip()
         spec['ip'] = ip
 
+        # Get 'mizar.com/egress-bandwidth' from pod annotations
+        egress_bw_bps = int(0)
+        annotations = self.param.body['metadata'].get('annotations', {})
+        if len(annotations) > 0:
+            k8s_egress_bw_bps = annotations.get(CONSTANTS.MIZAR_EGRESS_BW_TAG)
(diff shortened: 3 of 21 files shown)


kind-setup.sh
```diff
@@ -113,10 +113,20 @@ api_ip=`docker inspect -f '{{range .NetworkSettings.Networks}}{{.IPAddress}}{{en
...
+    fi
+    echo "Waiting for node ready.."
+    sleep 10
+done
+
 source install/create_crds.sh $CWD
 source install/create_service_account.sh $CWD
 
 source install/deploy_daemon.sh $CWD $MODE $DOCKER_ACC
+sleep 5
```
vinaykul: get rid of sleep 5

mizar/dp/mizar/workflows/builtins/pods/create.py
```diff
@@ -95,6 +95,21 @@ def run(self):
...
 
+        # Get 'mizar.com/egress-bandwidth' from pod annotations
+        egress_bw_bps = int(0)
+        annotations = self.param.body['metadata'].get('annotations', {})
+        if len(annotations) > 0:
+            k8s_egress_bw_bps = annotations.get(CONSTANTS.MIZAR_EGRESS_BW_TAG)
+            if k8s_egress_bw_bps is not None:
+                egress_bw_bps = k8s_egress_bw_bps
+                if k8s_egress_bw_bps.endswith('K'):
+                    egress_bw_bps = int(float(k8s_egress_bw_bps.replace('K', '')) * 1e3)
```
vinaykul: Should this be 1e3 or 1024? Need multiply by 8 on conversion?

src/xdp/trn_agent_xdp.c
```diff
@@ -233,12 +235,29 @@ static __inline int trn_encapsulate(struct transit_packet *pkt,
...
 	pkt->ip->ttl = pkt->inner_ttl;
 
+if (((pkt->inner_ipv4_tuple.saddr & 0xFF) == 20) && ((pkt->inner_ipv4_tuple.daddr & 0xFF) == 20)) {
+    if (pkt->inner_ipv4_tuple.protocol == IPPROTO_UDP) {
+	bpf_debug("[Agent:0x%x] VDBG EGRESS_BW: %lu INNER_TOS: %x!!!!!\n",
+		bpf_ntohl(pkt->agent_ep_ipv4), egress_bw_bps, pkt->inner_tos);
+	bpf_debug("[Agent:0x%x] VDBG PACKET_METADATA: POD_LABEL: %lu NS_LABEL: %lu\n",
+		bpf_ntohl(pkt->agent_ep_ipv4), pod_label_value, namespace_label_value);
+    }
+}
```
vinaykul: remove

src/xdp/trn_agent_xdp.c
```diff
@@ -415,7 +442,11 @@ static __inline int trn_process_inner_ip(struct transit_packet *pkt)
 			  __LINE__);
 	}
 
-	return trn_redirect(pkt, pkt->inner_ip->saddr, pkt->inner_ip->daddr);
+	int retcode = trn_redirect(pkt, pkt->inner_ip->saddr, pkt->inner_ip->daddr);
+
+bpf_debug("--------------ACTION=%d-------------------------------------\n\n",
```
vinaykul: remove

src/xdp/trn_agent_xdp.c
```diff
@@ -533,6 +564,8 @@ static __inline int trn_process_inner_eth(struct transit_packet *pkt)
 SEC("agent")
 int _agent(struct xdp_md *ctx)
 {
+bpf_debug("++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++\n"
```
vinaykul: remove

src/xdp/trn_edt_tc.c
```diff
@@ -0,0 +1,158 @@
...
+	char in_msg[] = "=====>>>>\n";
+	char out_msg[] = "<<<<===== [TC_Action: %d]\n";
+	char edt_msg[] = "[TC-EDT:0x%x] Low priority packet Dst: 0x%x\n";
+	void *data = (void *)(long)skb->data;
+	void *data_end = (void *)(long)skb->data_end;
+
+	//TODO: User creates this map
+	int key = 0;
+	struct edt_info einfo = {};
+	if (!bpf_map_lookup_elem(&THROTTLE_MAP, &key)) {
```
vinaykul: Clean this up. Configure from user-mode.

mizar/dp/mizar/workflows/builtins/pods/create.py
vinaykul: @Hong-Chang It looks like the current conversion is correct besides the need to multiply by 8 to convert it to bits per second (bps).  See https://en.wikipedia.org/wiki/Kilobyte
In kubernetes, the annotation such as "kubernetes.io/egress-bandwidth": "10M" refers to 10M/s (Megabytes per second = 1000x1000 bytes per second). The binary notation (Mi/s) would imply 1024x1024.

Hong-Chang: I see. Thanks.

Makefile
```diff
@@ -98,7 +98,10 @@ clean::
 	rm -f *.gcov
 
 .PHONY: test
-test:: lcov functest
+test:: test_docker_image lcov functest
+
+test_docker_image:
```
phudtran, line 103: We don't actually need this image anymore
It is a part of the old tests that is no longer used

phudtran (approved):
LGTM

Makefile
vinaykul: Ok. I'll roll this back in a follow-up PR later today - I need to make a few small corrections .. change location of map used for EDT to its own header file instead of using shared_map_defs.h as this is not shared with any of the other XDP code.

Should I get rid of the test target as well from the Makefile to avoid confusion? (We keep run_unittests)

vinaykul:
I'll address @phudtran issue https://github.com/CentaurusInfra/mizar/pull/492#discussion_r644493895 in a follow-up PR later today with some other changes.

PR #493, opened 2021-05-27
Implement Label Based Network Policies - Judge traffic pass or block

### LINKED ISSUE #457: Implement Label Based Network Policies - Judge traffic pass or block

Hong-Chang:
What does this PR do?

1. Break original label defined network policies
2. Retrieve label values from packet. (Move the function to more proper place)
3. Check from label-policy xdp table whether the rule allows the traffic based on label values
4. Checking policy is in "shortcut" mode. Pod, namespace, pod+namespace policy will be check sequentially. Will return once matching the policy.
5. When pod label changed, call update_packet_metadata

Why is it needed?
After the pr, the traffic will be judged pass or block by label based policy.
And this pr addresses an issue that there is no packet metadata change for pod update

How was this tested?
I did manual e2e test and the traffics were judged as expected.

No.

Closed #457


vinaykul:
Going forward, could you please add descriptive comments for commits? The description "update" is very generic and provides very little useful information about the change.

src/xdp/trn_transit_xdp.c
```diff
@@ -460,8 +460,50 @@ static __inline int trn_process_inner_ip(struct transit_packet *pkt)
...
+				return XDP_ABORTED;
+			}
+
+			if (pkt->namespace_label_value_opt->opt_class != TRN_GNV_OPT_CLASS) {
+				bpf_debug("[Scaled_EP:%d:0x%x] ABORTED: Unsupported Geneve option class\n",
+					__LINE__, bpf_ntohl(pkt->itf_ipv4));
+				return XDP_ABORTED;
+			}
+
+			if (pkt->namespace_label_value_opt->type != TRN_GNV_LABEL_VALUE_OPT_TYPE) {
```
vinaykul: I suggest changing this to two separate types: TRN_GNV_POD_LABEL_VALUE_OPT_TYPE & TRN_GNV_NAMESPACE_LABEL_VALUE_OPT_TYPE. This will clearly distinguish them when we capture them on the wire. And will help in the future when we figure out how to add only the types that we want to send instead of always bloating up the packet whether or not we have labels.

vinaykul:
I reviewed the changes in XDP code. Besides one minor comment, this looks good to me. 
We can merge it if @phudtran and @clu2xlu approve.

mizar/networkpolicy/networkpolicy_util.py
```diff
@@ -350,12 +350,12 @@ def fill_cidrs_from_directional_traffic_rules(self, data, policy_name, indexed_p
                             "pod_label_map": self.get_pod_label_values_by_label_dict(rule_item["podSelector"]["matchLabels"]),
                             "namespace_label_map": self.get_namespace_label_values_by_label_dict(rule_item["namespaceSelector"]["matchLabels"])
                         }
-
-                    namespaces = kube_list_namespaces_by_labels(networkpolicy_opr.core_api, rule_item["namespaceSelector"]["matchLabels"])
-                    if namespaces is not None:
-                        namespace_set = set()
-                        for namespace in namespaces.items:
-                            namespace_set.add(namespace.metadata.name)
+                    else:
```
clu2xlu, line 355: this else statement corresponding to if direction == "ingress", which is direction == "egress" right? ingress and egress rules should be similar. Basically please clarify the logic. Please add comments in network policy util.

mizar/networkpolicy/networkpolicy_util.py
```diff
@@ -366,23 +366,23 @@ def fill_cidrs_from_directional_traffic_rules(self, data, policy_name, indexed_p
...
-
-                    namespaces = kube_list_namespaces_by_labels(networkpolicy_opr.core_api, rule_item["namespaceSelector"]["matchLabels"])
-                    if namespaces is not None:
-                        for namespace in namespaces.items:
-                            pods = kube_list_pods_by_namespace(networkpolicy_opr.core_api, namespace.metadata.name)
-                            if pods is not None:
-                                for pod in pods.items:
-                                    self.add_pod_ip_into_cidrs_map(pod, policy_name, data["cidrs_map_no_except"][indexed_policy_name])
+                    else:
+                        namespaces = kube_list_namespaces_by_labels(networkpolicy_opr.core_api, rule_item["namespaceSelector"]["matchLabels"])
```
clu2xlu, line 374: same as above

mizar/networkpolicy/networkpolicy_util.py
```diff
@@ -366,23 +366,23 @@ def fill_cidrs_from_directional_traffic_rules(self, data, policy_name, indexed_p
...
                     self.add_label_networkpolicy(data, rule_item["podSelector"]["matchLabels"], policy_name)
                     if direction == "ingress":
                         data["pod_label_map"][indexed_policy_name] = self.get_pod_label_values_by_label_dict(rule_item["podSelector"]["matchLabels"])
-
-                    pods = kube_list_pods_by_labels(networkpolicy_opr.core_api, rule_item["podSelector"]["matchLabels"])
-                    if pods is not None:
-                        for pod in pods.items:                            
-                            self.add_pod_ip_into_cidrs_map(pod, policy_name, data["cidrs_map_no_except"][indexed_policy_name])
+                    else:
+                        pods = kube_list_pods_by_labels(networkpolicy_opr.core_api, rule_item["podSelector"]["matchLabels"])
```
clu2xlu, line 388: same

src/xdp/conntrack_common.h
```diff
@@ -93,6 +94,55 @@ static inline int enforce_ingress_policy(__u64 tunnel_id, const struct ipv4_tupl
 
 	if (0 == policies_ppo) return -1;
 
+	if (pod_label_value > 0) {
+		bpf_debug("Checking packet for pod label policy with pod_label_value=%d",
```
clu2xlu, line 98: connection tracking map only includes ip address. Can you explain the loggic over here? How are you applying label to connection tracking map?

src/xdp/trn_transit_xdp.c
```diff
@@ -460,8 +460,50 @@ static __inline int trn_process_inner_ip(struct transit_packet *pkt)
 				return XDP_ABORTED;
 			}
 		} else {
+			pkt->pod_label_value_opt = (void *)pkt->scaled_ep_opt + sizeof(*pkt->scaled_ep_opt);
```
clu2xlu, line 463: seems like you are forcing tracked_state to be null here for label based network policy. How do you know if the environment is label based or ip based? Consider separate this portion of the code from ip based network policy

clu2xlu (changes requested):

mizar/networkpolicy/networkpolicy_util.py
Hong-Chang: For ingress traffic, use label based policy. The label based policy only fits for ingress for now. For egress traffic, still keep original usage which is ip based policy. 
Added comments in the code to tell the difference.

mizar/networkpolicy/networkpolicy_util.py
Hong-Chang: For ingress traffic, use label based policy. The label based policy only fits for ingress for now. For egress traffic, still keep original usage which is ip based policy. 
Added comments in the code to tell the difference.

mizar/networkpolicy/networkpolicy_util.py
Hong-Chang: For ingress traffic, use label based policy. The label based policy only fits for ingress for now. For egress traffic, still keep original usage which is ip based policy. 
Added comments in the code to tell the difference.

src/xdp/conntrack_common.h
Hong-Chang: This file is conntrack_common.h but it contains function ingress_policy_check and inside function it calls enforce_ingress_policy. I have to adding the label based value check inside enforce_ingress_policy function.

It looks to me enforce_ingress_policy function is not in the right file because it's not "connection track". But such consideration was not from this release and I think we can refactor later.

src/xdp/trn_transit_xdp.c
Hong-Chang: There is only one environment. In the environment, when sending out the traffic, the code will always add label values to geneve header. And when receiving the traffic (the code here), the code will always retrieve the label values from geneve header.
The code here is in the function trn_process_inner_ip, which is contains by function trn_process_inner_eth, which is contains by function trn_process_geneve. The code is in "controlled" environment which rules the traffic must be from another side which is controlled by mizar. The both sides are implemented by mizar to use geneve to carry on additional flags for the traffic.

src/xdp/conntrack_common.h
clu2xlu: discussed with hong offline. resolved

clu2xlu (approved):

src/xdp/trn_transit_xdp.c
Hong-Chang: Splitted as suggested

Hong-Chang:
> Going forward, could you please add descriptive comments for commits? The description "update" is very generic and provides very little useful information about the change.

Thanks. These "update" actually from last pr (sending lable policy data to xdp table). I'm not adding more non-descriptive commits after that, but since this pr did "git fetch" and resolved conflicts with last pr, those commits are shown here.

phudtran (approved):
LGTM

PR #496, opened 2021-06-03
Network bandwidth limiting viaEDT - use bytes/sec and use node src IP as key in EDT eBPF map

vinaykul:
What does this change do?
This PR fixes egress bandwidth in Mizar code to use bytes per second instead of bits per second to align with the unit used by K8s annotation. 
It also uses the droplet host IP as key for the EDT rate-limiting BFP map.

Why is it needed?
This is a clean consistent way to manage bandwidth units, and use a key that can work for multiple NIC scenarios (when we do that in the future)

How was this tested?
Manual E2E testing. (Unit tests are already in with previous PR

No.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -98,10 +98,7 @@ clean::
 	rm -f *.gcov
 
 .PHONY: test
-test:: test_docker_image lcov functest
-
-test_docker_image:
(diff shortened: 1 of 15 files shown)


Hong-Chang (approved):
LGTM

phudtran (approved):
LGTM

PR #498, opened 2021-06-03
GitHub Actions CI, bug fixes, and tests

phudtran:
This PR fixes and adds the following:

- Adds CI via GitHub Actions
- Removes Travis CI (Travis org shutting down)
- Fixes multiple cluster bring-up issues with NoneType Error
- Fix for load transit CLI unit tests
- Adds new single deploy YAML
- Adds starter code for additional validation of operator errors in E2E tests
- Temporarily disables the following tests due to regressions/CI issues
  - Unit Test: Network Policy CLI 
  - E2E Tests: Scaled EP test (Regression)
  - E2E Tests: Simple EP test (Fails in GitHub Actions but not locally, more debugging needed)


etc/deploy/deploy.mizar.yaml
```diff
@@ -0,0 +1,492 @@
+---
```
vinaykul: Does this replace deploy.mizar.components.yaml file?

mizar/obj/tests/test.yaml
```diff
@@ -0,0 +1,62 @@
+apiVersion: v1
```
vinaykul, line 1: :)

vinaykul:
> * E2E Tests: Simple EP test (Fails in GitHub Actions but not locally, more debugging needed)

I took a quick scan and it looks good. Just one question about the failing action - will it block successful CI after this is merged?

/lgtm

phudtran:
> > * E2E Tests: Simple EP test (Fails in GitHub Actions but not locally, more debugging needed)
> 
> I took a quick scan and it looks good. Just one question about the failing action - will it block successful CI after this is merged?
> 
> /lgtm

Nope, I've disabled these for now, so they won't run

etc/deploy/deploy.mizar.yaml
phudtran: Yes it does, thanks for the catch
I will remove the old file and version the new docker images accordingly

vinaykul:
LGTM

Hong-Chang (approved):
LGTM

PR #513, opened 2021-07-07
Fix for deploying Arktos cluster with Mizar as default CNI plugin

click2cloud-team1:
**What type of PR is this?**
Fixes

- Using PR [#1114](https://github.com/CentaurusInfra/arktos/pull/1114), after Arktos cluster gets running, it takes a while to start the mizar-daemon, and once it gets running, we need to delete the mizar-operator pod for the first time. After CNI deployment, VPC and subnet are stuck in init state. Once the mizar-operator pod is recreated, all the network resources gets in provisioned state.
- Once we redeploy the cluster, all the network resources gets provisioned.
- This PR will resolve the above issue to avoid deletion of the mizar-operator pod for the first time.

### DIFF
--- a/etc/deploy/deploy.mizar.yaml
+++ b/etc/deploy/deploy.mizar.yaml
@@ -487,6 +487,16 @@ spec:
       serviceAccountName: mizar-operator
       terminationGracePeriodSeconds: 0
       hostNetwork: true
+      initContainers:
+      - name: daemon-init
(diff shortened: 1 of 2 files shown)


vinaykul:
We need to root-cause and understand why we have this issue in Arktos and not in upstream k8s.
/hold

vinaykul:
This change is not required.

PR #514, opened 2021-07-15
Configuration of eBPF maps for EDT bandwidth rate-limiting

vinaykul:
What does this change do?
This change is to allow configuration of rate-limiting bandwidth value into the eBPF maps that is used by the EDT rate-limiting eBPF program.
Why is it needed? 
This is part of the bandwidth QoS management feature that enables a user process to dynamically configure the limit for low priority traffic based on the monitored usage by high-priority traffic.
How was this tested?
Ping tests and traffic stream tesets.

### DIFF
--- a/mizar/daemon/interface_service.py
+++ b/mizar/daemon/interface_service.py
@@ -294,6 +294,16 @@ class InterfaceServer(InterfaceServiceServicer):
         rc, text = run_cmd(cmd)
         logger.info(
             "Disabled rx tx offload for host ep rc: {} text: {}".format(rc, text))
+
+        cmd = "nsenter -t 1 -m -u -n -i cat /sys/class/net/{}/speed".format(interface.veth.name)
+        rc, linkspeed = run_cmd(cmd)
+        linkspeed_bytes_per_sec = int(int(linkspeed.rstrip('\r\n')) * 100 * 100/ 8)
+        logger.info("Host interface {} Link Speed {} MB/s".format(interface.veth.name, linkspeed_bytes_per_sec))
+
+        #TODO: Get user-configured default bandwidth limit percentage from config-map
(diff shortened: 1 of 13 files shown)


mizar/daemon/interface_service.py
```diff
@@ -294,6 +294,16 @@ def ActivateHostInterface(self, request, context):
...
+
+        cmd = "nsenter -t 1 -m -u -n -i cat /sys/class/net/{}/speed".format(interface.veth.name)
+        rc, linkspeed = run_cmd(cmd)
+        linkspeed_bytes_per_sec = int(int(linkspeed.rstrip('\r\n')) * 100 * 100/ 8)
+        logger.info("Host interface {} Link Speed {} MB/s".format(interface.veth.name, linkspeed_bytes_per_sec))
+
+        #TODO: Get user-configured default bandwidth limit percentage from config-map
+        bwlimit = int((linkspeed_bytes_per_sec * CONSTANTS.MIZAR_DEFAULT_EGRESS_BW_LIMIT_PCT) / 100)
+        self.rpc.update_bw_qos_config(interface.address.ip_address, bwlimit)
+
```
phudtran, line 306: This runs on the main interface right?
If so it might be better to do it in daemon/app.py with all the other initialization steps.

Or is it actually for the host endpoint?

phudtran (approved):
Overall LGTM

mizar/daemon/interface_service.py
vinaykul: It's supposed to run on any interface that we create host endpoint for. If it were eth1 in a multi-nic scenario, then it should run on that. And if we create host-ep for multiple NICs, it should be attached to each of them. The idea is that we use source address as key, so each NIC gets its own EDT policing. 
Is app.py still the better place for it?

phudtran: Nope this is the place for it :)

vinaykul: Thanks. I'll merge it.

PR #518, opened 2021-07-26
Initial design specification for adding pods to a specific subnet/VPC

phudtran:
This PR specifies the interface for users to assign pods to specific VPCs and Subnets.

### DIFF
--- a/docs/design/mp_pod_to_vpc_subnet.md
+++ b/docs/design/mp_pod_to_vpc_subnet.md
@@ -1,0 +1,111 @@
+<!--
+SPDX-License-Identifier: MIT
+Copyright (c) 2021 The Authors.
+
+Authors: Phu Tran          <@phudtran>
+
+-->
+
+# Mizar User defined VPCs and Subnets for Vanilla Kubernetes
+
+## Introduction
+
+This document defines the interface for users to define their own VPCs and subnets. Furthermore, it also describes the interface for assigning pods to these specific subnet and VPC.
+
+## Custom VPC Creation
+
+When creating a VPC users must define the VPC's name, IP, prefix and the number of dividers.
+
+An example is given below.
+
+```yaml
+apiVersion: "mizar.com/v1"
+kind: Vpc
+metadata:
+  name: vpc0
+spec:
+  ip: "12.0.0.0"
+  prefix: "8"
+  dividers: 2
+```
+
+## Custom Subnet Creation
+
+When creating a subnet users must define the subnet's name, IP, prefix  the number of bouncers, and the VPC that it belongs to.
+
+An example is given below.
+
+```yaml
+apiVersion: "mizar.com/v1"
+kind: Subnet
+metadata:
+  name: net0
+spec:
+  vpc: "vpc1"
+  ip: "12.2.0.0"
+  prefix: "16"
+  bouncers: 2
+```
+The CIDR range the subnet is given must be within the range of the parent VPC's CIDR to be valid.
+
+## Pod to VPC and Subnet designation
+
+Three options are defined for assigning pods to specific VPCs.
+
+### Placing a pod into the default VPC
+
+Creating a standard pod with no Mizar specific annotation will assign it to the default VPC that is created at Mizar setup.
+
+```yaml
+apiVersion: v1
+kind: Pod
+metadata:
+  name: nginx
+spec:
+  containers:
+  - name: nginx
+    image: nginx
+    ports:
+      - containerPort: 443
+```
+
+### Placing a pod into a specific VPC
+
+Users can place a pod into a specific VPC by specifying the VPC annotation like below.
+
+```yaml
+apiVersion: v1
+kind: Pod
+metadata:
+  name: nginx
+  annotations:
+    mizar.com: {"vpc": "vpc0"}
+spec:
+  containers:
+  - name: nginx
+    image: nginx
+    ports:
+      - containerPort: 443
+```
+
+The pod is invalid if the VPC does not exist.
+
+### Placing a pod into a specific VPC and Subnet
+
+Users can place a pod into a specific VPC and subnet by specifying the required VPC and subnet annotation like below.
+
+```yaml
+apiVersion: v1
+kind: Pod
+metadata:
+  name: nginx
+  annotations:
+    mizar.com: {"vpc": "vpc0", "subnet": "subnet0"}
+spec:
+  containers:
+  - name: nginx
+    image: nginx
+    ports:
+      - containerPort: 443
+```
+The pod definition is invalid if the VPC or subnet does not exist.
--- a/docs/design/mp_pod_to_vpc_subnet.md
+++ b/docs/design/mp_pod_to_vpc_subnet.md
@@ -61,7 +61,7 @@ Three options are defined for assigning pods to specific VPCs.
 
 ### Placing a pod into the default VPC
 
-Creating a standard pod with no Mizar specific annotation will assign it to the default VPC that is created at Mizar setup.
+Creating a standard pod with no Mizar specific annotation will assign it to the default VPC and that VPC's default subnet that is created at Mizar setup.
 
 ```yaml
 apiVersion: v1
@@ -86,7 +86,7 @@ kind: Pod
 metadata:
   name: nginx
   annotations:
-    mizar.com: {"vpc": "vpc0"}
+    mizar.com/vpc: "vpc0"
 spec:
   containers:
   - name: nginx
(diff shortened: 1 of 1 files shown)


docs/design/mp_pod_to_vpc_subnet.md
```diff
@@ -0,0 +1,126 @@
...
+      - containerPort: 443
+```
+
+#### Requirements
+- The Parent VPC must exist.
+- The VPC must have an existing subnet.
+
+The pod is invalid if the VPC does not exist.
+
+### Placing a pod into a specific VPC and Subnet
```
pdgetrf, line 104: this may be beyond the scope of this design, but for workloads like deployment or job, user should also be able to specify vpc and subnet, and that could lead to more scenarios:

1. all pods from the same deployment belong to the same subnet
2. all pods from the same deployment belong to different subnets

docs/design/mp_pod_to_vpc_subnet.md
```diff
@@ -0,0 +1,126 @@
...
+
+Users can place a pod into a specific VPC and subnet by specifying the required VPC and subnet annotation like below.
+
+```yaml
+apiVersion: v1
+kind: Pod
+metadata:
+  name: nginx
+  annotations:
+    mizar.com: {"vpc": "vpc0", "subnet": "subnet0"}
```
pdgetrf: Inside XDP programs (for example, the transit XDP), I think VPCs and subnets are referenced/identified by VNI instead of names. In the  In the edge cluster situation, there will be cases where two subnets from two different clusters  belong to the same VPC. When a packet travels from one cluster to the other, VNI will be used in the process for matching, and if the VPC on one cluster has a different VNI from the VPC on the other cluster, it could drop the packet even though the VPCs on both clusters have the same name. In fact, there could also be a case where two VPCs in two cluster have the same name but with two different VNIs. In this case they are different VPCs. 

In summary, the VPC name -> VNI mapping works correctly in one cluster, but cross-cluster we will need to answer what is the unique identifier for a VPC? For example, if two VPC in different clusters have the same name but different VNI, is that considered the same VPC? What about different VPC names but the same VNI?

This needs to be answered or it will lead to complication down the road. I don't have an answer yet and we can discuss.

pdgetrf: I'd tent to think VPC name is the unique identifier cross-clusters. The different VNI could be converted by the edge gateway.

pdgetrf (commented):
left 2 questions.

docs/design/mp_pod_to_vpc_subnet.md
```diff
@@ -0,0 +1,126 @@
...
+
+An example is given below.
+
+```yaml
+apiVersion: "mizar.com/v1"
+kind: Vpc
+metadata:
+  name: vpc0
+spec:
+  ip: "12.0.0.0"
```
jshaofuturewei, line 27: I found there is no vni in subnet and vpc.  Do we still need vni or vni is not user defined any more?

docs/design/mp_pod_to_vpc_subnet.md
```diff
@@ -0,0 +1,126 @@
...
+- The Parent VPC must exist.
+- The new subnet must not yet already exist within the VPC.
+
+## Pod to VPC and Subnet designation
+
+Three options are defined for assigning pods to specific VPCs.
+
+### Placing a pod into the default VPC
+
+Creating a standard pod with no Mizar specific annotation will assign it to the default VPC that is created at Mizar setup.
```
jshaofuturewei: default subnet as well?

chenqianfzh:
> This PR specifies the interface for users to assign pods to specific VPCs and Subnets

Is it supported to assign an existing pod to a VPC/Subnet? Or update the VPC/Subnet of a running pod? I guess it is probably not for now. Could you please call it out in the design doc either it is supported or not?

docs/design/mp_pod_to_vpc_subnet.md
phudtran: VNI is no longer user defined.

docs/design/mp_pod_to_vpc_subnet.md
phudtran: Yes! Thanks for the catch. Will update

phudtran:
> > This PR specifies the interface for users to assign pods to specific VPCs and Subnets
> 
> Is it supported to assign an existing pod to a VPC/Subnet? Or update the VPC/Subnet of a running pod? I guess it is probably not for now. Could you please call it out in the design doc either it is supported or not?

That's right this is to supported for now. I will add this to the doc.

docs/design/mp_pod_to_vpc_subnet.md
```diff
@@ -0,0 +1,126 @@
...
+  dividers: 2
+```
+
+#### Requirements
+- The CIDR range is valid.
+- The VPC must not already exist.
+
+## Custom Subnet Creation
+
+When creating a subnet users must define the subnet's name, IP, prefix  the number of bouncers, and the VPC that it belongs to.
```
vinaykul, line 38: Can we consider supporting default values for number of dividers and bouncers? (Default 1, if not specified). Effectively, it would make these fields optional.

docs/design/mp_pod_to_vpc_subnet.md
```diff
@@ -0,0 +1,126 @@
...
+
+Users can place a pod into a specific VPC by specifying the VPC annotation like below.
+
+```yaml
+apiVersion: v1
+kind: Pod
+metadata:
+  name: nginx
+  annotations:
+    mizar.com: {"vpc": "vpc0"}
```
vinaykul: I'm not sure if this works. Annotations are usually specified as key: value.
e.g. mizar.com/vpc: vpc0

docs/design/mp_pod_to_vpc_subnet.md
```diff
@@ -0,0 +1,126 @@
...
+    image: nginx
+    ports:
+      - containerPort: 443
+```
+
+#### Requirements
+- The Parent VPC must exist.
+- The VPC must have an existing subnet.
+
+The pod is invalid if the VPC does not exist.
```
vinaykul, line 102: I'm not sure if it is possible to define a user-visible behavior for the pod state if this occurs. Can the operator update the pod status or write to the event stream?

docs/design/mp_pod_to_vpc_subnet.md
vinaykul: This would more likely be:
```golang
  annotations:
    mizar.com/vpc: vpc0
    mizar.com/subnet: subnet0
```

Please double-check if the above format is also supported for annotations.

docs/design/mp_pod_to_vpc_subnet.md
jshaofuturewei: I am thinking if we create pods with any existing vpcs or subnets and then create those vpcs and subnets, will we update those running pods into invalid states or just leave them alone?

docs/design/mp_pod_to_vpc_subnet.md
phudtran: Sure I can add that.

docs/design/mp_pod_to_vpc_subnet.md
phudtran: You're right will fix.

docs/design/mp_pod_to_vpc_subnet.md
phudtran: Agreed will fix.

docs/design/mp_pod_to_vpc_subnet.md
phudtran: Usually we retry a number of times before giving up in the operator. 
It would be preferable to just return in error to the user immediately, similar to syntax errors in a YAML file but I'm not sure if it's possible.

docs/design/mp_pod_to_vpc_subnet.md
phudtran: Agreed, I believe we can tackle this the same way as we do with pods.

If no VPC specified -> default VPC and default subnet for all pods.
If VPC specified but no subnet specified -> default subnets for all pods.

docs/design/mp_pod_to_vpc_subnet.md
pdgetrf: Discussed with @vinaykul  and @phudtran offline about this. The conclusion at the moment is to have the VPC name as the unique identifier of VPC. This means within a cluster, Mizar will generate VNI (VPC id) locally, and Edge Gateway will perform translation of VNI in and out of its cluster based on VPC name. VPC name needs to be global unique which will be enforced by the root level (cloud) cluster which all edge clusters connect with.

vinaykul: Sounds good. We'll merge this doc and work on this design. If we need to change something, we can always revisit.

vinaykul:
LGTM

PR #523, opened 2021-08-14
eBPF tail-call modules to record XDP tx stats, bandwidth monitor thread, and bw limit recommender algorithm, and multi-level network QoS for pods

vinaykul:
What does this change do?
This change adds XDP tail-call modules for Tx statistics recording. This will allow computation of high-priority traffic bandwidth usage and allow bandwidth-monitor thread (next PR) to configure the EDT rate-limit values for the TC eBPF program.

Why is it needed?
This is part of the bandwidth QoS management feature that enables high-priority traffic bandwidth computation.

How was this tested?
Ping tests when pinging from high-priority and low-priority pods and verifying the statistics updated in the tx_stats_t value field of tx_stats_map using bpftool.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -11,7 +11,6 @@ require (
 	github.com/vishvananda/netlink v1.1.1-0.20201029203352-d40f9887b852
 	golang.org/x/sys v0.0.0-20210630005230-0f9fa26af87c
 	google.golang.org/grpc v1.39.0
-	google.golang.org/grpc/cmd/protoc-gen-go-grpc v1.1.0 // indirect
-	google.golang.org/protobuf v1.26.0
+	google.golang.org/protobuf v1.27.0
 	k8s.io/klog/v2 v2.10.0
 )
--- a/go.mod
+++ b/go.mod
@@ -11,6 +11,7 @@ require (
 	github.com/vishvananda/netlink v1.1.1-0.20201029203352-d40f9887b852
 	golang.org/x/sys v0.0.0-20210630005230-0f9fa26af87c
 	google.golang.org/grpc v1.39.0
-	google.golang.org/protobuf v1.27.1
+	google.golang.org/grpc/cmd/protoc-gen-go-grpc v1.1.0 // indirect
+	google.golang.org/protobuf v1.26.0
 	k8s.io/klog/v2 v2.10.0
 )
--- a/mizar/dp/mizar/workflows/builtins/pods/create.py
+++ b/mizar/dp/mizar/workflows/builtins/pods/create.py
@@ -127,9 +127,11 @@ class k8sPodCreate(WorkflowTask):
 
         # Get 'mizar.com/egress-bandwidth' from pod annotations
         egress_bw = int(0)
-        pod_network_class_value = "BestEffort"
-        pod_network_priority_value = "Medium"
+        pod_network_class_value = "Premium"
+        pod_network_priority_value = "High"
         if os.getenv('FEATUREGATE_BWQOS', 'false').lower() in ('true', '1'):
+            pod_network_class_value = "BestEffort"
+            pod_network_priority_value = "Medium"
             annotations = self.param.body['metadata'].get('annotations', {})
             if len(annotations) > 0:
                 mizar_egress_bw = annotations.get(CONSTANTS.MIZAR_EGRESS_BW_TAG)
--- a/mizar/dp/mizar/workflows/builtins/pods/create.py
+++ b/mizar/dp/mizar/workflows/builtins/pods/create.py
@@ -127,24 +127,34 @@ class k8sPodCreate(WorkflowTask):
 
         # Get 'mizar.com/egress-bandwidth' from pod annotations
         egress_bw = int(0)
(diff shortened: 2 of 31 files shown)


src/dmn/trn_agent_xdp_usr.c
```diff
@@ -549,6 +553,51 @@ static int _trn_bpf_agent_prog_load_xattr(struct agent_user_metadata_t *md,
...
+	*prog_obj = NULL;
+	struct bpf_program *prog = NULL, *first_prog = NULL;
+	struct bpf_object *pobj = NULL;
+	struct bpf_map *pmap = NULL;
+	struct bpf_object_open_attr open_attr = { .prog_type = BPF_PROG_TYPE_XDP, .file = prog_file };
+
+	pobj = bpf_object__open_xattr(&open_attr);
+	if (IS_ERR_OR_NULL(pobj)) {
+		err = -PTR_ERR(pobj);
+		TRN_LOG_ERROR("Error openning bpf file: %s. err: %d:%s\n", open_attr.file, err, strerror(-err));
```
phudtran: "openning" typo :)

phudtran (approved):
LGTM

src/dmn/trn_agent_xdp_usr.c
vinaykul: thanks. fixed it.

go.mod
```diff
@@ -11,6 +11,7 @@ require (
 	github.com/vishvananda/netlink v1.1.1-0.20201029203352-d40f9887b852
 	golang.org/x/sys v0.0.0-20210630005230-0f9fa26af87c
 	google.golang.org/grpc v1.39.0
-	google.golang.org/protobuf v1.27.1
+	google.golang.org/grpc/cmd/protoc-gen-go-grpc v1.1.0 // indirect
+	google.golang.org/protobuf v1.26.0
```
Hong-Chang: I assume this change is necessary?

mizar/dp/mizar/workflows/builtins/pods/create.py
```diff
@@ -127,24 +127,36 @@ def run(self):
...
-                            float(k8s_egress_bw.replace('K', '')) * 1e3)
-                    elif k8s_egress_bw.endswith('M'):
+                            float(mizar_egress_bw.replace('K', '')) * 1e3)
+                    elif mizar_egress_bw.endswith('M'):
                         egress_bw = int(
-                            float(k8s_egress_bw.replace('M', '')) * 1e6)
-                    elif k8s_egress_bw.endswith('G'):
+                            float(mizar_egress_bw.replace('M', '')) * 1e6)
+                    elif mizar_egress_bw.endswith('G'):
                         egress_bw = int(
```
Hong-Chang, line 149: Looks to me 1e9 is near int upper range. Shall we do something for overflow when converting type to int?

src/cli/test/test_cli.c
```diff
@@ -1938,8 +1938,6 @@ static void test_trn_cli_update_packet_metadata_egress_bw_subcmd(void **state)
 	struct rpc_trn_packet_metadata_t exp_packet_metadata = {
 		.interface = "eth0",
 		.ip = 0x100000a,
-		.pod_label_value = 0,
-		.namespace_label_value = 0,
```
Hong-Chang, line 1942: Minor. I'm a little confused that here the label values are deleted (which is ok) but later in line 2107 the values are added. I think we can either need them in those places or remove them in those places.

go.mod
vinaykul: it showed up when i built after sync to latest. but it is not necessary. i'll remove it

mizar/dp/mizar/workflows/builtins/pods/create.py
vinaykul: i think the max int value is much bigger. there's lots of room considering we are most commonly talking 10G / 40G at this time.
```
root@ip-192-168-1-144:~/exvbwmizar# python
Python 2.7.17 (default, Feb 27 2021, 15:10:58) 
[GCC 7.5.0] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import sys
>>> print sys.maxint
9223372036854775807
>>>
```

src/cli/test/test_cli.c
vinaykul: This is intentional, I want the tests to exercise cases where the labels are missing and cases where they are present to ensure that my changes can co-exist with the label changes.

mizar/dp/mizar/workflows/builtins/pods/create.py
Hong-Chang: Ok, I checked doc and Python did special handling for int. I was concern the int defined to be max as 2,147,483,647, but Python will change int to long when necessary.

Hong-Chang (approved):
LGTM

go.mod
vinaykul: I reverted this change in my code. But this happens when you run bootstrap. Can you please check if the bootstrap script needs to be updated?

PR #524, opened 2021-08-16
Fix for containerd restart issue while deploying the Mizar YAML

click2cloud-team1:
- This PR fixes the containerd restart issue after applying the Mizar single deployment YAML.
- The conf file should be a regular file and it should be generated after the binary is placed in the respective location.
-  We have tested using the modified image of mizar container (c2cengg20190034/mizar:0.9) on OnPrem as well as GCE.


vinaykul:
I assume that you have tested it in GCE via kube-up.
Has this change also been tested for upstream k8s as well? Does this also work with arktos-up?

click2cloud-team1:
> I assume that you have tested it in GCE via kube-up.
> Has this change also been tested for upstream k8s as well? Does this also work with arktos-up?

Yes, we tested this via kube-up, arktos-up and in upstream k8s (kubeadm)

vinaykul (approved):
LGTM

phudtran (approved):

PR #526, opened 2021-08-18
Mizar cni convert from python to go

Hong-Chang:
What does this PR do?
Current mizar cni is written by python. As planned work now I convert it to go.

How was this tested?
I did e2e test. Automated test has been added. 

No.


vinaykul:
If not done already, can we please verify that it passes the Mizar health-check criteria? https://github.com/CentaurusInfra/mizar/wiki/Mizar-Cluster-Health-Criteria

.github/workflows/e2e_tests.yaml
```diff
@@ -55,6 +55,16 @@ jobs:
           git submodule update --init --recursive
       # - name: Update Kernel and Linux Headers
       #   run: printf 'y\nn\n' | sudo ./kernelupdate.sh
+      - name: Install go related
+        run: |
+          cd /tmp; wget https://dl.google.com/go/go1.13.9.linux-amd64.tar.gz
```
vinaykul: Why not use the latest verson of golang?

.github/workflows/e2e_tests.yaml
```diff
@@ -114,8 +127,8 @@ jobs:
           scp -r ~/actions-runner/_work/mizar/mizar ciw1:
           scp -r ~/actions-runner/_work/mizar/mizar ciw2:
 
-          ssh ciw1 "cd mizar && make"
-          ssh ciw2 "cd mizar && make"
+          ssh ciw1 "cd mizar && make container"
```
vinaykul: Should we not build the CNI driver as well each time we do make? This will likely break because the containers are built locally for these tests and any fixes to CNI won't be picked up on worker nodes.

vinaykul: @phudtran Please double-check me on this.

etc/docker/node-init.cni.sh
```diff
@@ -43,7 +43,7 @@ ln -snf /sys/fs/bpf /bpffs
 ln -snf /var/mizar/build/bin /trn_bin
 ln -snf /var/mizar/build/xdp /trn_xdp
 ln -snf /var/mizar/etc/cni/10-mizarcni.conf /etc/cni/net.d/10-mizarcni.conf
-ln -snf /var/mizar/mizar/cni.py /opt/cni/bin/mizarcni
+ln -snf /var/mizar/build/bin/mizarcni /opt/cni/bin/mizarcni
```
vinaykul: Can you please rebase to pick up latest commit? We need to use the C2C change in order for things to work for containerd. (PR https://github.com/CentaurusInfra/mizar/pull/524/files).

Also, this should be added to Dockerfile now that it is a static binary and should be installed to /opt/cni/bin

etc/docker/node-init.sh
```diff
@@ -63,7 +63,7 @@ if [[ "$parsedVersion" -lt "37" ]] ; then
 fi
 nsenter -t 1 -m -u -n -i mkdir -p /opt/cni/bin && \
 nsenter -t 1 -m -u -n -i mkdir -p /etc/cni/net.d && \
-nsenter -t 1 -m -u -n -i ln -snf /var/mizar/mizar/cni.py /opt/cni/bin/mizarcni && \
+nsenter -t 1 -m -u -n -i ln -snf /var/mizar/build/bin/mizarcni /opt/cni/bin/mizarcni && \
```
vinaykul: Same here. 

Why do we need both node-init.sh and node-init.cni.sh?

install/environment_adaptors/default_adaptor.sh
```diff
@@ -128,6 +128,6 @@ function install_dependencies_to_remote {
     kubectl exec $pod_name -- nsenter -t 1 -m -u -n -i ln -snf /var/mizar/build/bin /trn_bin
     kubectl exec $pod_name -- nsenter -t 1 -m -u -n -i ln -snf /var/mizar/build/xdp /trn_xdp
     kubectl exec $pod_name -- nsenter -t 1 -m -u -n -i ln -snf /var/mizar/etc/cni/10-mizarcni.conf /etc/cni/net.d/00-mizarcni.conf
-    kubectl exec $pod_name -- nsenter -t 1 -m -u -n -i ln -snf /var/mizar/mizar/cni.py /opt/cni/bin/mizarcni
+    kubectl exec $pod_name -- nsenter -t 1 -m -u -n -i ln -snf /var/mizar/build/bin/mizarcni /opt/cni/bin/mizarcni
```
vinaykul: cp static binary.

vinaykul (commented):
And has it been tested for both kind-setup.sh dev 3 and in real cluster using kubeadm?

k8s/kind/Dockerfile
```diff
@@ -25,7 +25,7 @@ RUN pip3 install /var/mizar/
 RUN ln -snf /var/mizar/build/bin /trn_bin
 RUN ln -snf /var/mizar/build/xdp /trn_xdp
 RUN ln -snf /var/mizar/etc/cni/10-mizarcni.conf /etc/cni/net.d/10-mizarcni.conf
-RUN ln -snf /var/mizar/mizar/cni.py /opt/cni/bin/mizarcni
+RUN ln -snf /var/mizar/build/bin/mizarcni /opt/cni/bin/mizarcni
```
vinaykul: cp static bin

.github/workflows/e2e_tests.yaml
Hong-Chang: I tried latest go1.17. From functionality it's working, but setup is a little different. For example, I see error "go get: installing executables with 'go get' in module mode is deprecated". After installed, then function it's good.

I think we need to decide a go version for now. go1.17 is current latest version, but sometime later it will be a normal "outdated" version. The reason I choose go1.13.9 is because it's the same go version in arktos. Since arktos and mizar both belong to Centaurus, I prefer this version. And When arktos changes version, mizar changes version too. Later after integration work of arktos and mizar, I don't hope we will hit issue because they are using different go version.

Hong-Chang:
> If not done already, can we please verify that it passes the Mizar health-check criteria? https://github.com/CentaurusInfra/mizar/wiki/Mizar-Cluster-Health-Criteria

Yes, I have done and it passes the criteria.

Hong-Chang:
> And has it been tested for both kind-setup.sh dev 3 and in real cluster using kubeadm?

Yes, has been tested in both environment.

.github/workflows/e2e_tests.yaml
Hong-Chang: CNI will be make in master machine and deploy to worker machines, so fix for CNI will be deployed to worker nodes. To make CNI, it requires to do some environment setup. I think it's reasonable to be done only in master machine. For node machine, maybe keep it as clean as possible then it can catch some real issue if it requires any dependency in worker node but we are not aware of.

etc/docker/node-init.cni.sh
Hong-Chang: It has been rebased. C2C change is respected.

etc/docker/node-init.sh
Hong-Chang: May be historic reason? @phudtran do you know? 

The both files are there so I updated both. I see the both files are being called so I think it's not a good idea to remove any of one.

install/environment_adaptors/default_adaptor.sh
Hong-Chang: I may not fully got your mean. Here /var/mizar/build/bin/mizarcni is not a directory. It's a binary file.

.github/workflows/e2e_tests.yaml
phudtran: These installation steps can be removed. Just install once on each of the machines and remove from the e2e tests. However, we should document these installation/setup steps in the bootstrap.sh file

etc/docker/node-init.sh
phudtran: node-init.cni.sh is old and not used. We can remove it

.github/workflows/e2e_tests.yaml
phudtran: Since you added make container to make all below, I believe you can just run make instead of `make && make container.`
An improvement suggestion, I think we can just run make once on the master and just copy the mizar folder over to the workers. Right now, it copies over the mizar folder first, then compiles on each machine.

phudtran (changes requested):
Added some comments

.github/workflows/e2e_tests.yaml
Hong-Chang: It's not "make && make container". It's "cd mizar && make container".
As discussed, we can improve the make process later. I created issue https://github.com/CentaurusInfra/mizar/issues/530 for tracking.

.github/workflows/e2e_tests.yaml
Hong-Chang: I tried remove and had thought it should be working but turns out it's not. I reconsidered this. Now I think our strategy could be "add more auto, reduce manual". When in any reason we need new CI machines, ideally we just need a fresh machine and the script will handle all the setup. I know that's not possible because kernel  needs to be updated, but we can try reduce manual work as less as we can. 
This go installation script doesn't costs much time I think it's no problem for now.

etc/docker/node-init.sh
Hong-Chang: There is a row of usages for node-init.cni. Please confirm all of them could be removed:
1. node-init.cni.sh is invoked by install-mizarcni.sh
2. install-mizarcni.sh is invoked by mizarcni.Dockerfile
3. mizarcni.Dockerfile is referenced by common.sh, line 172, to build mizarcni image
4. I'm not sure mizarcni docker image has usage or not.

install/environment_adaptors/default_adaptor.sh
vinaykul: Please see C2C PR https://github.com/CentaurusInfra/mizar/pull/524/files
This is not link anymore. It should be static binary.

k8s/kind/Dockerfile
vinaykul: https://github.com/CentaurusInfra/mizar/pull/524/files

This is not link anymore. It should be static binary.

vinaykul:
Please do the changes Phu and I have requested. I think they will help cleanup Mizar quite a bit by getting rid of dead code.

.github/workflows/e2e_tests.yaml
vinaykul: That's ok we can stick with go1.13,9 for now

.github/workflows/e2e_tests.yaml
vinaykul: I think it's better to build everything on the dev system and cp to worker nodes. That way we can avoid this container build target altogether and keep things simple.

vinaykul: The main reason I prefer to do cp is we avoid different container image ids, and if any local library idiosyncracies get picked up that can make it hard to debug. Let's do it right since we have the time.

Hong-Chang: ok, will do.

install/environment_adaptors/default_adaptor.sh
Hong-Chang: Yes I did a close to C2C pr. They copy for config file, but still link for binary. There are links in the same file. I feel it's risky to change them all. But I'm hesitate to change one which lead to different styles.

k8s/kind/Dockerfile
Hong-Chang: Yes I did a close to C2C pr. They copy for config file, but still link for binary. There are links in the same file. I feel it's risky to change them all. But I'm hesitate to change one which lead to different styles.

.github/workflows/e2e_tests.yaml
phudtran: Strange, why would we need to reinstall go every run?
Yes in the case we need new CI machines, it would be preferable to have them in the bootstrap.sh script rather than have them run every time in tests.

etc/docker/node-init.sh
phudtran: Yes we can remove the following files, and remove line 172 from common.sh

1. install-mizarcni.sh
2. mizarcni.Dockerfile
3. node-init.cni.sh

.github/workflows/e2e_tests.yaml
Hong-Chang: Done as requested.

.github/workflows/e2e_tests.yaml
Hong-Chang: I moved go install script from e2e test to bootstrap.sh.

etc/docker/node-init.sh
Hong-Chang: Removed.

phudtran (approved):
Looks good!

install/environment_adaptors/default_adaptor.sh
Hong-Chang: I changed to static binary.

vinaykul (approved):

PR #529, opened 2021-08-25
Fix mizar service issue: kubernetes service not working

Hong-Chang:
What does this PR do?
Currently in mizar environment, kubernetes service is not working. One behavior is that coredns pods are not ready. 
It's because kubernetes builtin endpoint object has no annotations hence not triggered event to add endpoint data into bpf map. This change is to add necessary annotation for the endpoint.

How was this tested?
I did e2e test. Now kubernetes service is working and coredns pods are ready.

No.


phudtran (changes requested):
Can you enable the scaled endpoint test for the CI to run for this change?
`teste2e/DISABLED_test_basic_scaled_endpoint.py`
Just remove the "DISABLED" prefix.

phudtran (approved):
LGTM!

PR #534, opened 2021-08-31
 Fix for bootstrap.sh script to update kernel headers to the latest kernel if ver>=5.6.0

### LINKED ISSUE #507: Fix bootstrap script to update kernel headers to the latest kernel if ver>=5.6.0
Currently, if the kernel is older than 5.6.0 bootstrap script will invoke kernel-update that updates kernel as well as the header.
However, for already updated kernels, the update is skipped, and headers are not updated causing a build break. Todo: Update kernel headers if required to match kernel version if ver>=5.6.0

click2cloud-team4:
**What type of PR is this?**

bug fix

**What this PR does / why we need it:**
This PR will Update kernel headers if kernel version >5.6.0

**Which issue(s) this PR fixes:**
Fixes #507 

**How was this tested?:**
Manual Tested:
we have tested this PR on **GCE** using arktos-up and kind-setup on following kernel version

- v5.6 to v5.13, Mizar pods are in running state ,CRDs are in provision state.
- Pod deployment is created successfully and able to ping each other.

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -11,6 +11,7 @@ logout_needed=false
 
 sudo apt-get update
 sudo apt-get install -y \
+sudo apt-get install linux-headers-$(uname-r)\
     build-essential \
     clang-7 \
     llvm-7 \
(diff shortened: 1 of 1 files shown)


bootstrap.sh
```diff
@@ -66,23 +66,42 @@ fi
 git submodule update --init --recursive
 
 kernel_ver=`uname -r`
+kernel_hed=`sudo apt search linux-headers-$(uname -r)`
+
 echo "Running kernel version: $kernel_ver"
+echo "..................................."
+echo "Running kernel headers: $kernel_hed"
```
vinaykul: Please use sensible naming. Rename this to kernel_header_ver

bootstrap.sh
```diff
@@ -66,23 +66,42 @@ fi
...
+
 mj_ver=$(echo $kernel_ver | cut -d. -f1)
 mn_ver=$(echo $kernel_ver | cut -d. -f2)
+mj_hed=$(echo $kernel_hed | cut -d. -f7 | cut -d- -f3)
```
vinaykul: mj_header_ver and mn_header_ver.

bootstrap.sh
```diff
@@ -66,23 +66,42 @@ fi
...
-	echo "Mizar requires an updated kernel: linux-5.6-rc2 for TCP to function correctly. Current version is $kernel_ver"
-
-	read -p "Execute kernel update script (y/n)?" choice
-	tput sgr0
-	case "$choice" in
-	  y|Y ) sh ./kernelupdate.sh;;
-	  n|N ) echo "Please run kernelupdate.sh to download and update the kernel!"; exit;;
-	  * ) echo "Please run kernelupdate.sh to download and update the kernel!"; exit;;
-	esac
+if ([[ "0$mj_ver" -le "04" ]] || [[ "$mn_ver" -le "05" ]]) || ([[ "$mn_ver" -le "5" ]]); then
```
vinaykul: What is "0$mj_ver"? And why do we need this change?

vinaykul (changes requested):

bootstrap.sh
click2cloud-team4: Updated the script with proper variable names, thanks.

bootstrap.sh
click2cloud-team4: Updated the variable names, thanks.

bootstrap.sh
click2cloud-team4: Previously the condition was not getting satisfied without the concatenation of '0', but now, we have optimized the condition's logic and the zero is removed and the code is updated.

vinaykul:
This PR does not work. apt search command is not the right command to use to list installed packages, rather it is used for searching available packages. Also the code to check for version should really be part of kernel update.

This PR is rejected due to lack of thoughtfulness in the changes being proposed, due to various quality issues and from history the general difficulty in getting any meaningful work done from Click2Cloud contributors. We will fix this issue ourselves.

PR #536, opened 2021-09-03
Development YAML

phudtran:
What does this change do?
Adds yaml for in development features.
Updates CRD definition to work for Kubernetes v1.16+.
Removes deprecated CRD definitions that no longer work in Kubernetes  v1.22+.

Why is it needed?
For users to try out in development features on real Kubernetes clusters outside of the Kind environment.

How was this tested?
Manually on a Kubeadm cluster


w-yue:
//LGTM
This is the same as the one I modified and have been using in my current benchmark testing. So I consider it's been tested at my side.

w-yue:
LGTM

vinaykul (approved):

Issue #485, opened 2021-05-18
XDP program fails to load when benchmark=True is set in Mizar kind-setup deployment


vinaykul:
To repro this issue, do the following:
1. In mizar codebase, set benchmark=true in mizar/common/rpc.py and mizar/daemon/interface_service.py
2. Deploy a two-worker kind cluster: './kind-setup dev 2'
3. Inspect the XDP programs loaded on veth-hostep@eth-hostep in the kind-worker nodes. This would be missing.
4. Inspect the mizar-daemon logs. It contains message such as "bpf failed to load program permission denied"
5. The above does not occur when running with DEBUG enabled (i.e. ebp_debug.o programs are loaded)

This should work regardless of whether _ebpf.o or ebpf_debug.o program binaries are used.

vinaykul:
I think we can consider this issue fixed. Wei: If there's nothing else left to do for this issue, could you please close it? thx

w-yue:
This issue is fixed in https://github.com/CentaurusInfra/mizar/pull/537. Close it now.

PR #537, opened 2021-09-15
Fix for Issue #485 https://github.com/CentaurusInfra/mizar/issues/485

w-yue:
What does this change do?
This change fixes Issue 485(XDP program fails to load when the benchmark flag is set to True in Mizar kind-setup deployment).

Why is it needed?
We need to be able to turn off debug mode for performance test and in general.

How was this tested?
Tested locally in aws kubeadm 2 node cluster setup. The XDP program was loaded successfully with benchmark=True. Also verified that there's around 30% performance difference with Benchmark=true in tested netperf and ping commands.

### DIFF
--- a/src/xdp/trn_agent_xdp.c
+++ b/src/xdp/trn_agent_xdp.c
@@ -62,9 +62,10 @@ static __inline int trn_select_transit_switch(struct transit_packet *pkt,
 	*s_port = SPORT_MIN + (inhash % SPORT_BASE);
 
 	if (*s_port < SPORT_MIN || *s_port > SPORT_MAX) {
-		bpf_debug("[Agent:] trn_select_transit_switch (BUG): s_port [%d] "
-			  "is outside of port range [%d - %d]!\n",
-			  *s_port, SPORT_MIN, SPORT_MAX);
+		bpf_debug("[Agent:%ld.0x%x] trn_select_transit_switch (BUG): s_port[%d] "
+			  " is outside of port range!\n",
+			  pkt->agent_ep_tunid, bpf_ntohl(pkt->agent_ep_ipv4),
+			  *s_port);
 		return 1;
 	}
 
@@ -126,9 +127,10 @@ static __inline int trn_encapsulate(struct transit_packet *pkt,
 		__u32 inhash = jhash_2words(in_src_ip, in_dst_ip, INIT_JHASH_SEED);
 		s_port = SPORT_MIN + (inhash % SPORT_BASE);
 		if (s_port < SPORT_MIN || s_port > SPORT_MAX) {
-			bpf_debug("[Agent:] DROP (BUG): s_port [%d] "
-			 			"is outside of port range [%d - %d]!\n",
-			  		s_port, SPORT_MIN, SPORT_MAX);
+			bpf_debug("[Agent:%ld.0x%x] DROP (BUG): s_port[%d] "
+			 		" is outside of port range!\n",
+					pkt->agent_ep_tunid, bpf_ntohl(pkt->agent_ep_ipv4),
+			  		s_port);
 			return XDP_DROP;
 		}
 
--- a/src/xdp/trn_agent_xdp.c
+++ b/src/xdp/trn_agent_xdp.c
@@ -62,6 +62,9 @@ static __inline int trn_select_transit_switch(struct transit_packet *pkt,
 	*s_port = SPORT_MIN + (inhash % SPORT_BASE);
 
 	if (*s_port < SPORT_MIN || *s_port > SPORT_MAX) {
+		bpf_debug("[Agent:] trn_select_transit_switch (BUG): s_port [%d] "
+			  "is outside of port range [%d - %d]!\n",
+			  *s_port, SPORT_MIN, SPORT_MAX);
 		return 1;
 	}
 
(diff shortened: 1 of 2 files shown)


src/xdp/trn_agent_xdp.c
```diff
@@ -62,6 +62,9 @@ static __inline int trn_select_transit_switch(struct transit_packet *pkt,
 	*s_port = SPORT_MIN + (inhash % SPORT_BASE);
 
 	if (*s_port < SPORT_MIN || *s_port > SPORT_MAX) {
+		bpf_debug("[Agent:] trn_select_transit_switch (BUG): s_port [%d] "
```
vinaykul: We should keep "Agent:%ld.0x%x" to distinguish node, vpc. There should be enough args left. If not, can this be condensed?

src/xdp/trn_agent_xdp.c
```diff
@@ -410,29 +422,29 @@ static __inline int trn_process_inner_ip(struct transit_packet *pkt)
 			__u8 conn_allowed = (pkt->inner_ipv4_tuple.protocol == IPPROTO_UDP) ? FLAG_REEVAL : 0;
 			conntrack_set_conn_state(&conn_track_cache, pkt->agent_ep_tunid, &pkt->inner_ipv4_tuple, conn_allowed);
 		}
-	}
 
-	/* Check if we need to apply a forward flow update */
+		/* Check if we need to apply a forward flow update */
```
vinaykul, line 428: Can you please fix this so tab use is consistent?

src/xdp/trn_agent_xdp.c
```diff
@@ -410,29 +422,29 @@ static __inline int trn_process_inner_ip(struct transit_packet *pkt)
...
-		/* Modify the inner packet accordingly */
-		trn_set_src_dst_port(pkt, out_tuple->sport, out_tuple->dport);
-		trn_set_src_dst_inner_ip_csum(pkt, out_tuple->saddr,
-					      out_tuple->daddr);
-		trn_set_dst_mac(pkt->inner_eth, out_tuple->h_dest);
-	} else {
-		bpf_debug("[Agent:%ld.0x%x] No dest IP address found! [%d]\n",
-			  pkt->agent_ep_tunid, bpf_ntohl(in_tuple.daddr),
-			  __LINE__);
+		if (out_tuple) {
```
vinaykul, line 437: Are these just due to tab change?

test/trn_controller/droplet.py
```diff
@@ -16,7 +16,7 @@
 
 
 class droplet:
-    def __init__(self, id, droplet_type='docker', phy_itf='eth0', benchmark=False):
+    def __init__(self, id, droplet_type='docker', phy_itf='eth0', benchmark=True):
```
vinaykul: We don't want to do this just yet.

vinaykul (changes requested):
Please see my comments.

src/xdp/trn_agent_xdp.c
w-yue: This is for tab consistency.

src/xdp/trn_agent_xdp.c
w-yue: Yes.

test/trn_controller/droplet.py
w-yue: Thought about that. Will switch back.

src/xdp/trn_agent_xdp.c
w-yue: This is a copy/paste from L129. I'll address both places.

w-yue: Actually L81 as well.

test/trn_controller/droplet.py
w-yue: switched back.

src/xdp/trn_agent_xdp.c
```diff
@@ -120,6 +124,16 @@ static __inline int trn_encapsulate(struct transit_packet *pkt,
...
+		__u32 inhash = jhash_2words(in_src_ip, in_dst_ip, INIT_JHASH_SEED);
+		s_port = SPORT_MIN + (inhash % SPORT_BASE);
+		if (s_port < SPORT_MIN || s_port > SPORT_MAX) {
+			bpf_debug("[Agent:%ld.0x%x] DROP (BUG): s_port[%d] "
+			 		" is outside of port range!\n",
+					pkt->agent_ep_tunid, bpf_ntohl(pkt->agent_ep_ipv4),
+			  		s_port);
+			return XDP_DROP;
+		}
+
```
phudtran, line 136: Looks like this could be causing a failure in the HTTP/TCP tests. Can you rerun the tests to see if it's transient?

phudtran (changes requested):

src/xdp/trn_agent_xdp.c
w-yue: I reran the e2efunctest and there's no complaints. Is this the one you're talking about?

phudtran: No, under the "Checks" tab of the PR, you can rerun all tests on the CI

w-yue: Just reran it, seems all passed.

phudtran (approved):
LGTM

vinaykul (approved):

PR #539, opened 2021-09-16
Fix issue 538 Bootstrap bash script failures

### LINKED ISSUE #538: Bootstrap bash script failures
The [bootstrap.sh](https://github.com/CentaurusInfra/mizar/blob/dev-next/bootstrap.sh) script got failed on [line-76 ](https://github.com/CentaurusInfra/mizar/blob/e6a1f88e2f25848c64ffd5f85c5c060cdec6f681/bootstrap.sh#L76) 

On line-76, the working directory is `/tmp` which is not a git repo, and running `git submodule update --init --recursive
` will result in the following error

```
fatal: not a git repository (or any of the parent directories): .git
```

### Steps to reproduce the error:
On fresh VM, clone mizar repo and run bootstrap.sh file 

***Output***
image

Hong-Chang:
What does this PR do?
In the original script there is a "cd /tmp" command which brings consequence to later script.  The fix is to remove the cd command and side effect will gone.

How was this tested?
I ran the issue script in a new machine and succeed.

No.

Closes #538

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -64,9 +64,9 @@ if [ ! -f /usr/local/bin/kind ]; then
 fi
 
 # Install go and related
-cd /tmp; wget https://dl.google.com/go/go1.13.9.linux-amd64.tar.gz
-sudo tar -C /usr/local -xzf go1.13.9.linux-amd64.tar.gz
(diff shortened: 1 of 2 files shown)


vinaykul (approved):
LGTM

PR #541, opened 2021-09-29
Fix udp service instable issue

Hong-Chang:
What does this PR do?
Current udp service is instable. Normally every 1 or 4 udp traffic throw service will be dropped due to checksum mismatch. I did investigation and figured out the issue happens in trn_update_l4_csum_port when updating port value for checksum. Its defined data type __be16 is leading issue when doing ~ operation. The type should be defined as __be32.

How was this tested?
I ran e2e test. Originally 1 or 4 traffic will drop. Now I tested hundreds of traffic and all succeed.

No.

### DIFF
--- a/src/xdp/trn_kern.h
+++ b/src/xdp/trn_kern.h
@@ -254,8 +254,8 @@ static inline void trn_update_l4_csum(__u64 *csum, __be32 old_addr,
 }
 
 __ALWAYS_INLINE__
-static inline void trn_update_l4_csum_port(__u64 *csum, __be16 old_port,
-					   __be16 new_port)
+static inline void trn_update_l4_csum_port(__u64 *csum, __be32 old_port,
+					   __be32 new_port)
 {
 	*csum = (~*csum & 0xffff) + ~old_port + new_port;
 	*csum = trn_csum_fold_helper(*csum);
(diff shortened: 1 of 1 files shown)


w-yue (commented):
LGTM

src/xdp/trn_kern.h
```diff
@@ -254,8 +254,8 @@ static inline void trn_update_l4_csum(__u64 *csum, __be32 old_addr,
 }
 
 __ALWAYS_INLINE__
-static inline void trn_update_l4_csum_port(__u64 *csum, __be16 old_port,
-					   __be16 new_port)
+static inline void trn_update_l4_csum_port(__u64 *csum, __be32 old_port,
+					   __be32 new_port)
```
vinaykul, line 258: ports are 16 bit numbers, why do we need to change it to 32 bit?

vinaykul:
There are lots of checksum failures for TCP. I don't think this fixes the issue. Please try TCP iperf.

src/xdp/trn_kern.h
vinaykul: I'll check this in for now as I don't see any obvious regressions and based on your testing that verified it fixes the instability for service connectivity. But I think we need to rewrite the checksum code. It is not trivial (except for the fix I made earlier) and I don't want to hold up the release for it.

vinaykul (approved):
I'll check this in for now as I don't see any obvious regressions and based on your testing that verified it fixes the instability for service connectivity. But I think we need to rewrite the checksum code. It is not trivial (except for the fix I made earlier) and I don't want to hold up the release for it.

PR #546, opened 2021-10-04
Zero-initialize UDP checksum (it is optional) in encapsulating GENEVE packet

vinaykul:
What does this PR do?
It sets the UDP checksum field in the encapsulating GEVENE outer packet to 0. This is optional and if zero, ignored on receipt. This prevents unexpected errors or drops.


Hong-Chang (approved):
LGTM

PR #553, opened 2021-10-20
Print EDT BPF trace messages only in debug mode

vinaykul:
In EDT eBPF program, the debug print trace is enabled by default. This affects performance benchmarking. 

The change fixes this so that only the debug version of the eBPF program binary has tracing enabled.


phudtran (approved):
LGTM

Issue #519, opened 2021-08-04
Add Mizar support for predictable interface naming


vinaykul:
Currently, Mizar assumes that the interface name of primary interface in a single NIC system is "eth0". This is not true in the newer OSes that use naming such as "ens33" etc..

We need to fix the Mizar control plane code to handle this scenario.

vinaykul:
We need this for the 9/30 milestone so Wei will work on fixing this issue.

click2cloud-ashwin:
Hi @w-yue 

Recently I was working on this issue. I came to know that @vinaykul Sir assigned this issue to you. I have made some findings which may help you to resolve this issue.

I will share my internal public repo link. I hope, you will find it useful.

Kindly revert if any confusion arises.

w-yue:
This issue is closed with solution https://github.com/CentaurusInfra/mizar/pull/555.

PR #555, opened 2021-10-28
Fix for Issue #519(Add Mizar support for predictable interface naming)

w-yue:
What does this PR do?
We currently uses hard-coded "eth0" as nic interface in our code which prevents Mizar from running in system whose default interface is not named as eth0, which is the case in the later Ubuntu version or some other Linux flavors.

This PR remove hard coded "eth0" from the logic and replaces with the default interface name obtained from the system. It is a fix for issue #519.

How was this tested?
It passes local kind e2efunctest and it's also tested in an AWS 2 nodes cluster environment with default interface name as "ensxx", the k8s cluster was launched successfully and the default interface name is detected and used in Mizar pods(instead of eth0), the testpods were created successfully and ping-able to each other.

No.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,7 +44,8 @@ CFLAGS += -Wall
 CFLAGS += -Wextra
 CFLAGS += -Werror
 CFLAGS += -pedantic -Wpedantic
-CLFAGS += -Wno-cast-function-type -Wno-error=cast-function-type
+# needed in 20.04
+#CFLAGS += -Wno-cast-function-type -Wno-error=cast-function-type
 
 CFLAGS += -fno-common
 CFLAGS += -fstrict-aliasing
--- a/mizar/common/common.py
+++ b/mizar/common/common.py
@@ -442,8 +442,17 @@ def get_default_itf():
             return line.split()[4]
     return default_itf
 
+def get_bridge_real_itf():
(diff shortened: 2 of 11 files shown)


vinaykul (approved):
LGTM

Makefile
```diff
@@ -44,7 +44,8 @@ CFLAGS += -Wall
 CFLAGS += -Wextra
 CFLAGS += -Werror
 CFLAGS += -pedantic -Wpedantic
-CLFAGS += -Wno-cast-function-type -Wno-error=cast-function-type
+# needed in 20.04
+#CFLAGS += -Wno-cast-function-type -Wno-error=cast-function-type
```
phudtran, line 48: Is there a way to fix this for 20.04 without removing these flags?
I believe there should be some warnings that pop up during compilation.

phudtran (approved):
Overall LGTM. Thanks Wei!

Makefile
w-yue: Sorry little misleading here. There's actually no effect for this change. This flag was never effective due to previous typo(was **CLFAGS** instead of **CFLAGS**). I added it back in 20.04 with typo fix but it won't work for 18.04 so commented it out.

PR #559, opened 2021-11-02
Fix issue 552 Creating, Deleting, and once again creating Pod does not work with Mizar

### LINKED ISSUE #552: Creating, Deleting, and once again creating Pod does not work with Mizar
Repro steps:

1. Create a 1-master 1-worker kubernetes cluster with kubeadm.
2. Deploy mizar as the network provider via by creating deploy.mizar.yaml from mizar repo.
3. Create a simple pod and verify that it is up and running.
4. Delete the pod.
5. Create the same pod again - you will find it stuck in ContainerCreating state.

Hong-Chang:
What does this PR do?
When deleting pod, the code needs to execute "ip netns delete" against name of network namespace. The issue is that the code should send a name as parameter, but currently it's sending the file name together with folder name. Hence it's failed to delete netns, then pod cannot be created again.
The fix is to remove the folder name, then the netns will be deleted.

How was this tested?
I run the fixed binary in cluster and verified the fix is working. Now the pod can be created, deleted, and then created again without issue.

No.

### DIFF
--- a/mizar/daemon/interface_service.py
+++ b/mizar/daemon/interface_service.py
@@ -236,7 +236,7 @@ class InterfaceServer(InterfaceServiceServicer):
         """
         veth_peer_index = get_iface_index(interface.veth.peer, self.iproute)
         self.rpc.unload_transit_agent_xdp(interface)
-        self.iproute.link('del', index=veth_peer_index)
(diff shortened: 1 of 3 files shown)


vinaykul (approved):
LGTM
Thanks.

PR #560, opened 2021-11-03
Print EDT eBPF program trace messages only in debug mode

vinaykul:
In EDT eBPF program, the debug print trace is enabled by default. This affects performance benchmarking.

The change fixes this so that only the debug version of the eBPF program binary has tracing enabled.

Also temporarily disabled QoS feature-gate in kind-setup until I've had the chance to figure out intermittent test regression.


phudtran (approved):
LGTM

PR #566, opened 2021-11-11
Don't annotate endpoints when annotate has been done

Hong-Chang:
What does this PR do?
mizar will update annotation for endpoint object. But when the object has the exact same annotation, it's not necessary to add it. This PR checks annotation and ignore the action.
The PR will breaks the dead lock of keeping adding the annotation which fails arktos-mizar deployment

Also add a deploy.mizar.next.yaml for team to deploy latest mizar image. We cannot use deploy.mizar.dev.yaml since it's for the kubernetes 1.22.

How was this tested?
e2e tested

No.


phudtran (approved):
LGTM

PR #569, opened 2021-11-23
Fix for duplicate Arktos pod requests

phudtran:
This PR fixes an issue with pod creation and networking for the Arktos integration

### DIFF
--- a/mizar/arktos/arktos_service.py
+++ b/mizar/arktos/arktos_service.py
@@ -38,9 +38,10 @@ store = OprStore()
 
 class ArktosService(BuiltinsServiceServicer):
 
-    def CreatePod(self, request, context):
+    def ProcessPodRequest(self, request):
+        return_code = None
         if request.host_ip == '':
-            return ReturnCode(
+            return_code = ReturnCode(
                 code=CodeType.TEMP_ERROR,
                 message="Missing hostIP during pod create or update"
             )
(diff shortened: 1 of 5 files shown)


mizar/arktos/arktos_service.py
```diff
@@ -52,17 +55,26 @@ def CreatePod(self, request, context):
         param.body['status']['phase'] = request.phase
         param.body['metadata']['tenant'] = request.tenant
         param.extra = {}
+        ep_name = request.name+"-default-{}-eth0".format(request.tenant)
```
Hong-Chang: "default" and "eth0" may be subjected to be variable?

mizar/arktos/arktos_service.py
```diff
@@ -52,17 +55,26 @@ def CreatePod(self, request, context):
...
+        if ep_opr.store.get_ep(ep_name):
+            logger.info("EP {} already exists! Return OK".format(ep_name))
+            return ReturnCode(
```
Hong-Chang: I think pod create and update are sharing same code. So if user wants to change pod by purpose, such as change its priority, seems to me the change will be unexpectedly ignored.

mizar/daemon/interface_service.py
```diff
@@ -81,13 +85,13 @@ def _CreateVethInterface(self, interface):
         veth_peer = interface.veth.peer
         veth_index = get_iface_index(veth_name, self.iproute)
 
-        if veth_index != -1:
-            self.iproute.link('delete', index=veth_index)
-            veth_index = -1
         if veth_index == -1:
+            logger.info("Creating if {}".format(veth_name))
```
Hong-Chang: if => interface

mizar/arktos/arktos_service.py
phudtran: Right. The interface part may be anything. However, I'm always seeing that "default". Let me dig into this a bit more.

mizar/daemon/interface_service.py
phudtran: Fixed

mizar/arktos/arktos_service.py
phudtran: Fixed

mizar/arktos/arktos_service.py
phudtran: Fixed.

Hong-Chang (approved):
LGTM assuming the change has been tested.

PR #575, opened 2021-12-02
Set default traffic QoS class/priority in agent eBPF program to premium/high.

vinaykul:
What does this PR do?
This changes the default traffic classification for egress traffic QoS classification in agent XDP program to premium/high so that uses the XDP_REDIRECT path as default. This is a potential fix for an issue in Mizar-Arktos integration.
There's also a minor cleanup change.

How was this tested?
kind-setup and ping tests.

No.


Hong-Chang (approved):
LGTM

PR #577, opened 2021-12-07
Use premium/high as default DSCP code in agent XDP/eBFP program.

vinaykul:
Fix the default DSCP code in the agent XDP program to premium/high to preserve the behavior that existed before Pod QoS feature was implemented.

Tested with multi-node K8s cluster and Kind cluster - tested for CRDs provisioning and ping connectivity.


Hong-Chang (approved):
LGTM

PR #581, opened 2021-12-11
Onboarding Arktos Ci

phudtran:
This PR adds the initial work for Arktos-Mizar tests in CI.

### DIFF
--- a/install/common.sh
+++ b/install/common.sh
@@ -55,15 +55,27 @@ function common:check_cluster_ready {
 # Checks for status: Provisioned for given object
 function common:get_object_status {
     local object=$1
+    local cluster=$2
 
-    kubectl get $object 2> ${KUBECTL_LOG} | awk '
-    NR==1 {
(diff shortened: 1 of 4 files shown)


install/common.sh
```diff
@@ -87,11 +100,17 @@ function common:check_mizar_ready {
 
 function check_all_ready {
     timeout=360
+    local cluster="kubernetes"
```
vinaykul, line 103: Do you think we should have check_mizar_ready as part of arktos-up (when CNIPLUGIN==mizar)?

phudtran: I believe we would have to copy over this function to arktos-up.sh. Unfortunately we can't just call this function as arktos-up.sh deploys mizar via yaml, and doesn't have access to the mizar repo's scripts.

vinaykul: yeah let's keep it here for now to  get this moving. we have to add it to arktos-up and we can do that later.

vinaykul (approved):

PR #585, opened 2022-01-06
Set more timeout for grpc

Hong-Chang:
What does this PR do?
This pr change cni grpc timeout from 3 sec to 8 sec. The reason is that in mizar/daemon/interface_service.py, function ConsumeInterfaces, there is a while loop which timeout in 5 seconds. So in grpc client it shall expect some cases happens over 5 seconds. Now set grpc timeout to 8 sec.

During arktos-mizar integration, we observed many cases that pod cannot be running. The issue is grpc failed with "rpc error: code = DeadlineExceeded desc = Deadline Exceeded" which is grpc timeout. This pr will fix the issue.

How was this tested?
Tested in e2e environment

No.

### DIFF
--- a/pkg/util/grpcclientutil/grpcclientutil.go
+++ b/pkg/util/grpcclientutil/grpcclientutil.go
@@ -63,7 +63,7 @@ func getInterfaceServiceClient() (InterfaceServiceClient, *grpc.ClientConn, cont
 		return nil, nil, nil, nil, err
 	}
 	client := NewInterfaceServiceClient(conn)
-	ctx, cancel := context.WithTimeout(context.Background(), time.Second*3)
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*8)
 	return client, conn, ctx, cancel, nil
 }
 


phudtran (approved):
LGTM

pkg/util/grpcclientutil/grpcclientutil.go
```diff
@@ -63,7 +63,7 @@ func getInterfaceServiceClient() (InterfaceServiceClient, *grpc.ClientConn, cont
 		return nil, nil, nil, nil, err
 	}
 	client := NewInterfaceServiceClient(conn)
-	ctx, cancel := context.WithTimeout(context.Background(), time.Second*3)
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*8)
```
vinaykul, line 66: Can we please add this to constants file as CONTEXT_TIMEOUT and reference it here? 
Also, does it make sense to take a larger number (perhaps 30s)?

vinaykul (commented):
Please see comment

pkg/util/grpcclientutil/grpcclientutil.go
Sindica: Does it make sense for this call to keep retry?

PR #590, opened 2022-01-11
Fix issue 588: Once pod creating failed, it will fail all the way

### LINKED ISSUE #588: Once pod creating failed, it will fail all the way
**What observed:**
When failed to creating a pod from a deployment, it will fail all the way. From cni log (/tmp/mizarcni.log) it shows k8s/arktos keeps retrying "Del", "Add" operations to the pod, but pod cannot start.
At this time, if using "kubectl delete pod" to delete the pod, since it's a deployment, a new pod will start and running good.

**The issue:**
When creating a pod, it creates an endpoint. Endpoint will invoke ProduceInterface. When cni does "Add" operation, cni will invoke ConsumeInterface. Eventually interface is produced and consumed, and the pod is running.
When cni sends "Del" operation, currently DeleteInterface will be invoked. It destroys objects from both ProduceInterface and DeleteInterface. So next time when cni sends "Add" operation, it will always fail because ProduceInterface is only invoked once and it's not there any more.

**Solution:**
DeleteInterface shall not belong to cni operation. It shall only be invoked when deleting endpoint. In this way, ProduceInterface and DeleteInterface will both be invoked once. Cni retrying "Del", "Add" operations won't affect the pod creation.

Hong-Chang:
**What type of PR is this?**
>
/kind bug

**The issue:**
When creating a pod, it creates an endpoint. Endpoint will invoke ProduceInterface. When cni does "Add" operation, cni will invoke ConsumeInterface. Eventually interface is produced and consumed, and the pod is running.
When cni sends "Del" operation, currently DeleteInterface will be invoked. It destroys objects from both ProduceInterface and DeleteInterface. So next time when cni sends "Add" operation, it will always fail because ProduceInterface is only invoked once and it's not there any more.

**Solution:**
Move DeleteInterface to only be invoked when deleting endpoint. In this way, ProduceInterface and DeleteInterface will both be invoked once. Cni retrying "Del", "Add" operations won't affect the pod creation.

Fixes #588 

No

### DIFF
--- a/mizar/store/operator_store.py
+++ b/mizar/store/operator_store.py
@@ -168,6 +168,8 @@ class OprStore(object):
                 ep.egress_networkpolicies = old_ep.egress_networkpolicies
             if len(old_ep.data_for_networkpolicy) > 0 and len(ep.data_for_networkpolicy) == 0:
                 ep.data_for_networkpolicy = old_ep.data_for_networkpolicy
+            if old_ep.interface is not None and ep.interface is None:
+                ep.interface = old_ep.interface
 
         # logger.info('caller name:{}'.format(inspect.stack()[1][3]))
         self.eps_store[ep.name] = ep
--- a/cmd/mizarcni/app/worker.go
+++ b/cmd/mizarcni/app/worker.go
@@ -99,12 +99,6 @@ func DoCmdDel(netVariables *object.NetVariables, stdinData []byte) (string, erro
 	}
 	strBuilder.WriteString(fmt.Sprintf("Network variables: %s", netVariables))
 
-	strBuilder.WriteString(fmt.Sprintf("\nDeleting interface for %s/%s", netVariables.K8sPodNamespace, netVariables.K8sPodName))
-	err := grpcclientutil.DeleteInterface(*netVariables)
-	if err != nil {
-		return strBuilder.String(), err
-	}
-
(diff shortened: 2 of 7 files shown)


vinaykul:
Looks good to me. I assume that interface cleanup works the same as before and there's nothing left behind when pod is deleted. 
Could you also please try the scenario where the same pod yaml is created-deleted two or more times?

vinaykul (approved):

mizar/store/operator_store.py
```diff
@@ -168,6 +168,8 @@ def update_ep(self, ep):
                 ep.egress_networkpolicies = old_ep.egress_networkpolicies
             if len(old_ep.data_for_networkpolicy) > 0 and len(ep.data_for_networkpolicy) == 0:
                 ep.data_for_networkpolicy = old_ep.data_for_networkpolicy
+            if old_ep.interface is not None and ep.interface is None:
```
phudtran, line 171: `if old_ep.interface and not ep.inteface`

phudtran (approved):
LGTM. One comment

Hong-Chang:
> Looks good to me. I assume that interface cleanup works the same as before and there's nothing left behind when pod is deleted. Could you also please try the scenario where the same pod yaml is created-deleted two or more times?

Yes, I tried that.

PR #593, opened 2022-01-17
Use directory mount instead of file mount for optional KUBECONFIG in Mizar operator

vinaykul:
**What type of PR is this?**
>
> /kind design
> /kind feature
/kind bug
> /kind cleanup
> /kind documentation

**What this PR does / why we need it**: For scaleup and upstream, we should in_cluster_config. kubelet expects /etc/kubernetes/admin.conf file to exist even if it is not used when in_cluster_config is to be used. Using directory mount instead of file mount fixes the issue.

Fixes #

```release-note

```

### DIFF
--- a/etc/deploy/deploy.mizar.dev.yaml
+++ b/etc/deploy/deploy.mizar.dev.yaml
@@ -553,9 +553,9 @@ spec:
           privileged: true
         volumeMounts:
         - name: kubeconfig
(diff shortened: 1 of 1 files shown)


Hong-Chang (approved):
LGTM

phudtran (approved):

Issue #589, opened 2022-01-10
Slowness to start multiple pods concurrently


Hong-Chang:
What observed:
When using deployment to start 10 pods at the same time, it's slow for all the pods become running state.

The issue:
In code mizar/daemon/interface_service.py, function ConsumeInterfaces, there is code "while True:". When pods are being started concurrently, this part is inefficient. It keeps get pod name from queue, check whether it's desired pod, and put it back into the queue is it's not desired pod. When there are multi threads getting pod, comparing pod name, and putting back, the result is not determine and slow.

PR #601, opened 2022-01-26
Fix pod slowness, and Vanilla kubernetes Scaled Endpoint Creation

phudtran:
This PR address issue #589.
This also fixes an issue for vanilla Kubernetes with the scaled endpoints not being created due to wrong naming

### DIFF
--- a/mizar/daemon/interface_service.py
+++ b/mizar/daemon/interface_service.py
@@ -17,7 +17,6 @@ from google.protobuf import empty_pb2
 from mizar.common.rpc import TrnRpc
 from mizar.common.constants import *
 from pyroute2 import IPRoute, NetNS
-import queue
 import ipaddress
 
 logger = logging.getLogger('mizar_interface_service')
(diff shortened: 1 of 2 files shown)


Hong-Chang (approved):
LGTM

mizar/daemon/interface_service.py
```diff
@@ -135,14 +133,11 @@ def _ConsumeInterfaces(self, pod_name, cni_params):
...
-                self.queued_pods.remove(pod_name)
-            interfaces = self.interfaces.get(pod_name, [])
-            for interface in interfaces:
-                if interface.status == InterfaceStatus.queued:
-                    self._ProvisionInterface(interface, cni_params)
-                    interface.status = InterfaceStatus.consumed
+        interfaces = self.interfaces.get(pod_name, [])
+        for interface in interfaces:
+            if interface.status == InterfaceStatus.queued:
+                self._ProvisionInterface(interface, cni_params)
```
vinaykul, line 139: Does this need to be synchronized? I'm just wondering what the intent of the original mutex was.

mizar/dp/mizar/workflows/builtins/services/create.py
```diff
@@ -74,8 +74,10 @@ def run(self):
                             "VPC {} has no subnets to allocate service {}!".format(vpc_name, self.param.name))
                 net = net_opr.store.get_net(subnet_name)
         namespace = self.param.body['metadata']['namespace']
-        tenant = self.param.extra.get('tenant', '')
-        name = self.param.name + "-{}-{}".format(namespace, tenant)
+        name = self.param.name + "-{}".format(namespace)
+        if self.param.extra:
+            tenant = self.param.extra.get('tenant', '')
```
vinaykul, line 79: Is this non empty only in the arktos scenario?

vinaykul:
Overall LGTM. Just one minor clarification question to see if we didn't miss anything.

vinaykul (approved):

PR #602, opened 2022-01-27
Fix issue 599 ip address kept allocating

### LINKED ISSUE #599: When pod running issue and keeps restarting, ip addresses are kept allocating to the pod and wasted
**What happened**:
When there is a pod in the cluster, it has issue and keep restarting by cni, the ip will keep being allocated to the pod. The ip addresses are wasted in such scenario. We need to return ip back when pod starting failed.

**What you expected to happen**:
Ip resource shall only be used for running pods.

**How to reproduce it (as minimally and precisely as possible)**:
Start a cluster, assuming there is a pod has issue then keep restarting, or make a such pod. The allocatable ip addresses become less and less.

Hong-Chang:
For issue 599, we observed ip keeps being allocated for the same pod.
After more investigating, I found every time when pod got updated, cluster will trigger pod updated event and will be handled by mizar  k8sPodCreate workflow. And in the workflow the ip will be allocated.

In the workflow, there are two parts: part 1 will be run again and again each time when the pod updated. And part 2 will be run only once when pod is in "pending" state. Current ip allocating is in part 1, hence ip will keep allocating.

The fix is to move ip allocating from part 1 to part 2. In this pr, the ip allocating part is moved right after part 2 starts.

**What type of PR is this?**
>
> /kind design
> /kind feature
/kind bug
> /kind cleanup
> /kind documentation

Fixes #599 

```release-note

```

### DIFF
--- a/mizar/dp/mizar/workflows/builtins/pods/create.py
+++ b/mizar/dp/mizar/workflows/builtins/pods/create.py
@@ -112,11 +112,6 @@ class k8sPodCreate(WorkflowTask):
                 configs = json.loads(net_config)
                 spec['interfaces'] = configs
 
-        n = net_opr.store.get_net(spec['subnet'])
-        ip = n.allocate_ip()
-        logger.info("ip {} from {} is allocated to Pod {}.".format(ip, spec['subnet'], self.param.name))
-        spec['ip'] = ip
-
         # Get 'mizar.com/egress-bandwidth' from pod annotations
         egress_bw = int(0)
         pod_network_class_value = "Premium"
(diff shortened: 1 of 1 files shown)


phudtran (approved):
LGTM

mizar/dp/mizar/workflows/builtins/pods/create.py
```diff
@@ -170,6 +165,11 @@ def run(self):
             self.finalize()
             return
 
+        n = net_opr.store.get_net(spec['subnet'])
```
vinaykul, line 168: Thanks for figuring this out so fast.

I'm wondering if we should have a separate update.py file for on.update trigger, that will make it a lot cleaner to write future code. (not now - post 1/30 if we get to it)

vinaykul (approved):

PR #603, opened 2022-02-01
Fix system pod startup and enabled scaled ep tests

phudtran:
This PR does the following:

- Enable Scaled endpoint tests
- Fix an issue where system pods fail to start due to agent RPCs being called before the agent is loaded.


mizar/common/rpc.py
```diff
@@ -212,6 +212,8 @@ def update_agent_metadata(self, ep):
         cmd = f'''{self.trn_cli_update_agent_metadata} -i \'{itf}\' -j \'{jsonconf}\''''
         logger.info("update_agent_metadata: {}".format(cmd))
         returncode, text = run_cmd(cmd)
+        if "error" in text:
```
vinaykul: Is this the fix to catch the race condition where update hits before agent load?

phudtran: Yeah, the return code always returns none, but the output text will have "error" when something goes wrong.
I can turn this into a constant.

teste2e/common/k8s.py
```diff
@@ -147,7 +147,7 @@ def pod_exec_stdout(self, name, cmd):
                       stdout=True, tty=False)
         return resp
 
-    def create_service(self, name):
+    def create_service(self, name, vpc="vpc0", subnet="net0"):
```
vinaykul: Constants.py default_ep_vpc / default_ep_subnet ?

teste2e/common/k8sservice.py
```diff
@@ -20,6 +20,7 @@ def eptype(self):
     def backends(self):
         return set(self.endpoints.keys())
 
-    def add_endpoint(self, name):
-        ep = self.api.create_pod(name, scaledep=self.name)
+    def add_endpoint(self, name, vpc_name="vpc0", subnet_name="net0"):
```
vinaykul: same here. from constants default_ep_vpc/net?

vinaykul (approved):
A couple of minor fix suggestions.
LGTM.

teste2e/common/k8sservice.py
phudtran: Fixed

teste2e/common/k8s.py
phudtran: Fixed

PR #622, opened 2022-02-10
Fix issue 618 Recreated pod failed to be running

### LINKED ISSUE #618: Recreated pod failed to be running
There is a recent regression noticed. For a normal pod, create it and it will get running. Then delete it. Then create it again. At this time it cannot reach running state.

By a simple investigation we see the regression begins from commit 603. I checkout the code to the one right before the commit, and I don't see the issue. Then I checkout the code to commit 603, the issue can be repro'd.

**How to reproduce it (as minimally and precisely as possible)**:
1. Prepare the pod:
cat <<EOF | tee /tmp/testpod.yaml
apiVersion: v1
kind: Pod
metadata:
  name: testpod
spec:
  containers:
  - name: arktospoddefault1
    image: mizarnet/testpod
    ports:
      - containerPort: 443
EOF
2. Create the pod
cluster/kubectl.sh apply -f /tmp/testpod.yaml
3. Wait for pod to be running. 
4. Delete the pod
cluster/kubectl.sh delete pod testpod
5. Create the pod again
cluster/kubectl.sh apply -f /tmp/testpod.yaml

The pod should be able to be running but it won't.

Hong-Chang:
This pr is for issue 618 Recreated pod failed to be running. After consumed interface, the original code doesn't clean the wait list so has side impact to later request. This is to do the necessary list clean up. Tested the fix is working.

**What type of PR is this?**
/kind bug

Fixes #618

### DIFF
--- a/mizar/daemon/interface_service.py
+++ b/mizar/daemon/interface_service.py
@@ -145,6 +145,7 @@ class InterfaceServer(InterfaceServiceServicer):
 
         interfaces = InterfacesList(interfaces=interfaces)
         logger.info("Consumed {}".format(interfaces))
+        self.interfaces.pop(pod_name)
         return interfaces
 
     def _ProvisionInterface(self, interface, cni_params):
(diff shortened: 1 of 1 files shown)


phudtran (approved):
LGTM

vinaykul:
Thanks for fixing this!
/LGTM

PR #623, opened 2022-02-11
Use while to let ConsumeInterfaces waiting for ProduceInterfaces

Hong-Chang:
When starting a pod, currently it may need 15 seconds to be in running state. After this pr, for a single pod, it reduces to 2 seconds to be in running state.

**The reason behind:**
ConsumeInterface and ProduceInterface are from two different process. ConsumeInterface depends on ProduceInterface. If ConsumeInterface runs first, the process will fail and will be retried after 15 seconds, that's the reason we see the pod is running after 15 seconds.
This pr is to add a wait for ConsumeInterface until ProduceInterface has done. Although the code sets a timeout of 5 seconds, I see a single pod will be running in 2 seconds.

### DIFF
--- a/mizar/daemon/interface_service.py
+++ b/mizar/daemon/interface_service.py
@@ -29,6 +29,7 @@ logger.addHandler(handler)
 logger = logging.getLogger()
 
 CONSUME_INTERFACE_TIMEOUT = 5
+WAITING_SLEEP_INTERVAL = 0.5
 
 
 class InterfaceServer(InterfaceServiceServicer):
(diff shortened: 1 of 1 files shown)


mizar/daemon/interface_service.py
```diff
@@ -203,12 +203,20 @@ def ConsumeInterfaces(self, request, context):
...
+        # So using while here to wait until ProduceInterfaces has been done.
+        start = time.time()
+        while True:
+            if self.pod_dict:
+                if requested_pod_name in self.pod_dict:
+                    if self.pod_dict[requested_pod_name]:
+                        # Interfaces for the Pod has been produced
+                        self.pod_dict.pop(requested_pod_name)
+                        return self._ConsumeInterfaces(requested_pod_name, request)
+            time.sleep(0.5)
```
vinaykul: Can we please make the 0.5 a constant? (in the same file next to CONSUME_INTERFACE_TIMEOUT is good enough)

Hong-Chang: Const has been added.

vinaykul (approved):

phudtran (approved):

PR #625, opened 2022-02-17
Avoid spurious XDP program reloads and dup eBPF map table entries

vinaykul:
**What type of PR is this?**
>
> /kind design
> /kind feature
/kind bug
> /kind cleanup
> /kind documentation

**What this PR does / why we need it**: Load agent/transit XDP programs tries to initialize and load program even though it has already been loaded. This avoids creation of spurious eBPF map entries by avoiding load (without unload) if XDP program has already been loaded.

**Which issue(s) this PR fixes**: Arktos issues 1360.

Fixes #

**Special notes for your reviewer**: Tested scaleup with kube-up several times and scaleout 2x2.

```release-note

```

### DIFF
--- a/src/dmn/trn_rpc_protocol_handlers_1.c
+++ b/src/dmn/trn_rpc_protocol_handlers_1.c
@@ -605,7 +605,7 @@ error:
 int *load_transit_xdp_1_svc(rpc_trn_xdp_intf_t *xdp_intf, struct svc_req *rqstp)
 {
 	UNUSED(rqstp);
-	static int result = -1;
+	static int result = 0;
 
 	int rc;
 	bool unload_error = false;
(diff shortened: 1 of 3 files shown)


src/dmn/trn_rpc_protocol_handlers_1.c
```diff
@@ -605,7 +605,7 @@ rpc_trn_endpoint_t *get_ep_1_svc(rpc_trn_endpoint_key_t *argp,
 int *load_transit_xdp_1_svc(rpc_trn_xdp_intf_t *xdp_intf, struct svc_req *rqstp)
 {
 	UNUSED(rqstp);
-	static int result = -1;
+	static int result = 0;
```
Hong-Chang, line 608: This is changing logic and need to be careful whether there is any side effect.

Hong-Chang (approved):

phudtran (approved):
LGTM

src/dmn/trn_rpc_protocol_handlers_1.c
vinaykul: There isn't. The old code was setting it to 0 at the end on success path. This sets it to 0 initially and an approp error code is set before taking any error path.

PR #627, opened 2022-02-18
Host Endpoint create and gRPC fix

phudtran:
This PR fixes an issue where host endpoints were not being created for all VPCs
It also fixes an issue with all gRPC communication between the operator and the daemon for when the daemon is not yet ready.

### DIFF
--- a/mizar/daemon/app.py
+++ b/mizar/daemon/app.py
@@ -27,10 +27,22 @@ POOL_WORKERS = 10
 def init(benchmark=False):
     # Setup the droplet's host
     default_itf = get_itf()
+    script = (f''' bash -c 'for name in $(nsenter -t 1 -m -u -n -i ip l |  grep -Po "(vehost-\w+)(?=@)"); do nsenter -t 1 -m -u -n -i sudo ip l delete $name; done ' ''')
+    r = subprocess.Popen(script, shell=True, stdout=subprocess.PIPE)
+    output = r.stdout.read().decode().strip()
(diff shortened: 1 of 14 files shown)


mizar/daemon/app.py
```diff
@@ -27,10 +27,22 @@
...
+    script = (f''' bash -c 'for name in $(nsenter -t 1 -m -u -n -i ip l |  grep -Po "(veth-\w+)(?=@)"); do nsenter -t 1 -m -u -n -i sudo ip l delete $name; done' ''')
+    r = subprocess.Popen(script, shell=True, stdout=subprocess.PIPE)
+    output = r.stdout.read().decode().strip()
+    logging.info("Deleted preexisting veths: {}".format(output))
+
+    script = (f''' bash -c 'for file in $(nsenter -t 1 -m -u -n -i ls -1 /etc/cni/net.d/ | grep -v '10-mizarcni.conf$'); do nsenter -t 1 -m -u -n -i rm -rf /etc/cni/net.d/$file; done' ''')
+    r = subprocess.Popen(script, shell=True, stdout=subprocess.PIPE)
+    output = r.stdout.read().decode().strip()
+    logging.info("Deleted other CNI files: {}".format(output))
+
```
vinaykul, line 55: Do you think we should add code to cleanup /var/mizar and /opt/cni/bin/mizarcni files as well?

vinaykul (approved):

PR #645, opened 2022-03-08
Fix cluster startup errors

phudtran:
This PR fixes the following issues:

- Runtime errors where a dictionary changed size while being iterated through.
- An issue where the veth creation returned before the device was created.
- Timeout errors caused by the mizar-daemon taking longer than 15 minutes to come up in the Arktos kube-up environment.
- A runtime error where delete was called on a nonexistent key in a dictionary.
- An issue where pods deployed before mizar did not have network connectivity


vinaykul (approved):
LGTM. I didn't take a very close look but couldn't spot any obvious issues.

PR #646, opened 2022-03-11
added licence info

yanmo96:
**What type of PR is this?**
documentation

Added license info to files missing it

Fixes #

None

```release-note

```

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -2,6 +2,26 @@
 # Read the Docs configuration file
 # See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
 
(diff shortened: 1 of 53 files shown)


vinaykul (approved):
LGTM.
Thanks for working late and fixing this.
@phudtran Please take a second look. To me this looks consistent with what we currently have, but just confirming with Deepak if we can have two different licenses in the same project. AFAIK, the entire project becomes GPL if one or more files in the project are GPL.

phudtran (approved):
LGTM
Mostly just making sure anything using libraries from the Linux kernel is GPL.
Thanks Yan!

PR #660, opened 2022-09-01
Fix CNI driver to generate SPEC compliant response

vinaykul:
**What type of PR is this?**
>
> /kind design
> /kind feature
/kind bug
> /kind cleanup
> /kind documentation

**What this PR does / why we need it**: The current CNI implementation does not return CNI ADD response compliant to the CNI spec. This causes CNI ADD failure in newer versions of containerd that has stricter checking.

Fixes #

```release-note

```


phudtran (approved):
LGTM

PR #661, opened 2022-09-08
Override MAC address in Tx eBPF program for kubemark

vinaykul:
**What type of PR is this?**
>
> /kind design
/kind feature
> /kind bug
> /kind cleanup
> /kind documentation

**What this PR does / why we need it**: We use the MAC address of target host NIC for pod to pod communication. However this leads to filtering drops for kubemark case where the node is really a pod, being served by kubenet CNI. We need to keep the same behavior that as the host stack for populating destination MAC.

Fixes #

```release-note

```


phudtran (approved):
LGTM

